PR #3, opened 2015-06-18
Fixing javadoc for Java 8

kane77:
Java 8 has stricter javadoc checking (it cannot have <pre> nested in <code>, no '<' and '>' characters in javadoc and some stricter checking on @param names)


germanattanasio:
Hi @kane77, thanks for the fix. I was using a property to disable the strict javadoc checking from java 8:
```xml
<profile>
	<id>java8-disable-strict-javadoc</id>
	<activation>
		<jdk>[1.8,)</jdk>
	</activation>
	<properties>
		<javadoc.doclint.none>-Xdoclint:none</javadoc.doclint.none>
	</properties>
</profile>
```

Did you try to generate the javadoc without that property in the pom?

kane77:
Hi German,
after my change the generation of javadoc succeeds without the property in pom, as well as with gradle.
I guess the property could be removed from the pom.
Cheers, Martin

Issue #10, opened 2015-07-10
Add test cases


germanattanasio:
We need tests cases to make sure that code changes don't break the existing functionality. We need to choose a good framework to mock rest calls. 

I heard good comments about http://www.mock-server.com/

- [x] Add coveralls.io
- [x] Choose the test framework to use
- [x] Add test cases for the GA services

germanattanasio:
HI @alseddnm, can you help me with this?

germanattanasio:
fixed by @alseddnm

PR #19, opened 2015-08-20
Apimock

germanattanasio:
Merge  the `apimock` branch into dev. This includes the test case created by @alseddnm.


germanattanasio:
This is related to #10

PR #34, opened 2015-09-10
Add Retrieve and Rank service

gjs29:


germanattanasio:
this may require a different branch and some research

PR #67, opened 2015-10-02
Concept Insights: fix createCorpus

### LINKED ISSUE #66: Concept Insights: fix corpus id in createCorpus
Concept Insights API requires corpus `id` to be of the form `/corpora/{account_id}/{corpus}`. `createCorpus()` takes in a `Corpus` object, where an `id` is required. `createCorpus()` creates a URL for the request by concatenating `/v2/corpora/{account_id}` and the `id` from the `Corpus` object; currently this concatenation results in `/v2/corpora/{account_id}/corpora/{account_id}/corpus`.

Proposed solution: since Concept Insights API does not require an `id` for the corpus, allow `Corpus.id` to be `null` in `createCorpus()`. Add a `name` field to `Corpus` and use that when calling `createCorpusIdPath()` from `createCorpus()`.

msushkov:
Closes #66


germanattanasio:
@alseddnm can you take a look at this. It looks weird that we need another field like `name`. We probably want to change `createCorpusIdPath()` instead.

PR #73, opened 2015-10-12
Update Concept Insights models and tests cases based on a code review with nizar

germanattanasio:
* Update CI models to allow users to specify the `accountId` and `corpusName`
* Renamed `IDValidator` to `IDHelper`
* Splitted `getRelatedConcept` into two methods that now require parameters
outside of the parameters map
* updated test cases


germanattanasio:
This is related to #67

PR #78, opened 2015-10-13
renaming Insigths to Insights(typo fix)

fmerici:
Renamed dir path for Visual Insights as it had a typo.  Also fixed the package name in test class VisualInsightsTest.java to match new dir path

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ gradlew
 gradlew.bat
 target
 src/test/resources/config.properties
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -11,3 +11,4 @@ gradlew
 gradlew.bat
 target
 src/test/resources/config.properties
+bin/
```
germanattanasio: I forgot this...

PR #84, opened 2015-10-16
Added negative unit tests to VisualInsightsTest for bad requests

fmerici:
Added a couple of unit tests to test for bad requests, 400 error


germanattanasio:
Nice job @fmerici !

PR #89, opened 2015-10-26
Handles missing testing properties safely

colindean:
Checks for nulls and empty strings in properties when retrieved, then throws a rather explicit exception when something is wrong.

Backstory:

I checked out the code to use the R&R branch, but didn't understand what was wrong when I ran tests and got a bunch of NPEs. I dug a little more and found that there's a developer-specific configuration file necessary. Put the file in place, but still couldn't figure out what goes into it. This was a result, as I could easily find all of the properties needed through this and through some grepping.

A sample config is forthcoming.

### DIFF
--- a/src/test/java/com/ibm/watson/developer_cloud/MultiThreadTest.java
+++ b/src/test/java/com/ibm/watson/developer_cloud/MultiThreadTest.java
@@ -86,10 +86,10 @@ public class MultiThreadTest extends WatsonServiceTest{
 		super.setUp();
 		service = new LanguageTranslation();
 		service.setUsernameAndPassword(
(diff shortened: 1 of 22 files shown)


colindean:
I've got a commit ready for @kshasan's master branch that includes R&R, but there's a bunch of merge conflicts that I'm not familiar enough with the project to wade through. I'll hold off on submitting that until his branch merges/rebases this repo's master.

colindean:
Test failure appears to be in the setup of the test in an `openssl` generation step.

germanattanasio:
Thanks for the changes, next time make the PR to `dev` not `master`

colindean:
:+1:

PR #90, opened 2015-10-26
Adds a sample config.properties

colindean:
Actually added before #89, but sufficient to be in a separate commit.


build.gradle
```diff
@@ -2,8 +2,8 @@ apply plugin: 'java'
 apply plugin: 'maven'
 apply plugin: 'eclipse'
 
-sourceCompatibility = 1.6
-targetCompatibility = 1.6
+sourceCompatibility = 1.7
+targetCompatibility = 1.7
```
colindean: Crap, this shouldn't be here :-1: 

But I needed 1.7 in order for it to compile...

PR #99, opened 2015-11-10
variable uses List<Feed> but a wrapper class Feeds with the List inside exists

ruslan120101:


germanattanasio:
I'm doing this manually because the merge has conflicts

PR #100, opened 2015-11-10
Please verify, but seems Concept inherits from the wrong class

ruslan120101:


ruslan120101:
Please keep me updated if the **Concept** class should indeed inherit from **AlchemyLanguageGenericModel**. I'll be accordingly updating the iOS SDK.

ruslan120101:
@germanattanasio ? Please see comments above ^

germanattanasio:
That's fixed already. I'll be pushing the changes later today

germanattanasio:
Now that I'm looking I think concept should be inheriting from any class.

`EnrichedTitle` and `CombinedResults` are classes that use `Concept` and those are extending `AlchemyLanguageGenericModel` already

germanattanasio:
The class should be:
```
public class Concept {
```

PR #127, opened 2015-11-30
Adds a forgotten field

ruslan120101:


germanattanasio:
I'm looking at the [documentation](http://www.alchemyapi.com/api/authors-web-api) and there is no `confident` there. 

The `DocumentAuthors` object only used in the `AlchemyLanguage.getAuthors()` call.

ruslan120101:
I've observed occasional inconsistencies between the documentation, [Swagger](https://watson-api-explorer.mybluemix.net/apis/alchemy-language-v1), and the actual returned object.

In cases where the three differ would you recommend we go by the JavaSDK? In this case I went by the returned object from Alchemy.

Tagging @vherrin to keep everyone in the loop

ruslan120101:
(In this case Alchemy returns a "confident" object as well)

germanattanasio:
@ruslan120101 can you give me a curl command to test that?

ruslan120101:
Easy

```
curl -X POST --header "Content-Type: application/json" --header "Accept: application/json" "https://watson-api-explorer.mybluemix.net/alchemy-api/calls/url/URLGetAuthors?url=www.google.com"
```

germanattanasio:
Cool, now the problem will be to map the `no` to `false` which I don't think it's automatic. I will take a look at this tomorrow.

ruslan120101:
Great. I made it a **String** for now elsewhere

germanattanasio:
I've updated the model with a String. Thanks @ruslan120101 for not only communicating the problem but also fixing it!

PR #129, opened 2015-12-02
DCS: Adds support for using custom configs

mfulgo:
Adds SDK support for passing custom configurations to the Document Conversion Service. Configuration is passed as a JsonObject in order to minimize SDK changes in the event of future configuration changes.

### DIFF
--- a/src/main/java/com/ibm/watson/developer_cloud/document_conversion/v1/DocumentConversion.java
+++ b/src/main/java/com/ibm/watson/developer_cloud/document_conversion/v1/DocumentConversion.java
@@ -13,13 +13,6 @@
  */
 package com.ibm.watson.developer_cloud.document_conversion.v1;
 
-import java.io.File;
-import java.io.IOException;
-import java.io.InputStream;
-import java.util.Map;
-import java.util.logging.Level;
-import java.util.logging.Logger;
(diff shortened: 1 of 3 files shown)


germanattanasio:
@mfulgo I don't think that a `JsonObject` is a good solution. It's simple to implement in the SDK but it will not be easy for users to write code. I guess the problem here is that you want to keep configuration as generic as possible.

Can you please write the code using a non-trivial config with the current implementation?

I want to see the proposed solution in action before taking a decision. I still think that having a `ConversionOptions` is the right way to do it.

mfulgo:
@germanattanasio That's correct, we want to maximize the flexibility of the configuration and not have to re-create model objects if we introduce new config parameters or change/remove existing ones. *For now*, I think this is the best approach; if users find it difficult to work with, we can always enhance it later.

> Can you please write the code using a non-trivial config with the current implementation?

I'm not sure what you're looking for: a non-trivial config could just be a more complicated/larger version of the `custom_config.json` file.

nfriedly:
I'm merging this because I know we're pressed for time and need to ship it. But, for the record, I agree with @germanattanasio than a custom config object would be preferable over the JsonObject one here.

PR #130, opened 2015-12-02
Use Strings to parse Solr cluster size in Retrieve and Rank

gjs29:
The Retrieve and Rank service uses a String tp represent a Solr cluster's
size in its JSON responses. As a result, our Gson objects have to also
use Strings otherwise serialization will fail.


germanattanasio:
I don't think this is really a problem but I do feel that we should make the String size to be Integer. The change will be released next weekend

gjs29:
The problem that I overlooked earlier when I changed it back to Integer is that our service returns JSON as either {"cluster_size" : "2"} or {"cluster_size" : ""}. In the first case Gson can handle converting the String to an Integer. But in the second case Gson doesn't know what to do and throws this exception:
com.google.gson.JsonSyntaxException: java.lang.NumberFormatException: empty String
So the client doesn't work for free clusters at the moment.

As far as the SolrClusterOptions object goes, users only have to deal with the Integer form with this commit, so the String internal doesn't leak into that API here. The internal String field is what were stuck with though until we change the actual RR service.

germanattanasio:
great! thanks for the explanation

PR #136, opened 2015-12-07
Use preemptive Basic auth when making request with SolrJ

gjs29:
This prevents requests from being repeated when attempting
to authenticate through DataPower which causes SolrJ to fail.


germanattanasio:
We are sending credentials and the cookies and datapower is authenticating the request two times ?

gjs29:
This only affects the SolrJ stuff. Here's an example of what this failure looks like: http://grokbase.com/t/lucene/solr-user/1288xjjbwx/http-basic-authentication-with-httpsolrserver

PR #145, opened 2015-12-15
Added Timestamp / Word Confidence / Confidence Serialization/Deserialization to STT 

PranavSathy:
Added GSON serializer/deserializers for the timestamps, word_confidence & confidence attribute within the SpeechAlternative class. This needs to be integration tested, as my config.properties was empty and therefore had no ability to make API calls.


PranavSathy:
It's way too late at night... I need to go to sleep, sorry for the several opens :/

PR #146, opened 2015-12-15
Added Timestamp / Word Confidence / Confidence Serialization/Deserialization to STT 

PranavSathy:
Added GSON serializer/deserializers for the timestamps, word_confidence & confidence attribute within the SpeechAlternative class. This needs to be integration tested, as my config.properties was empty and therefore had no ability to make API calls.

PR #148, opened 2015-12-17
[speech-to-text] Add support for Timestamp and Word Confidence

PranavSathy:
Added support for Timestamps and Word Confidences utilizing GSON Annotations and TypeAdapter as per @germanattanasio suggestion in PR #146 .

Still need to run Integration Testing.


germanattanasio:
@PranavSathy were you able to test it locally?

PranavSathy:
@germanattanasio Not quite, I only have a free license for speech-to-text, it passed unit tests but I cannot run the integration-tests, I will look into trying this locally for just STT

PR #153, opened 2016-01-09
Upgraded Visual Recognition from V1 to V2

josephlam1984:
Hello German,

Based on the migration guide here http://www.ibm.com/smarterplanet/us/en/ibmwatson/developercloud/doc/visual-recognition/migrating.shtml, and the information from the API explorer, I've updated the Visual Recognition source code in the SDK from V1 to V2.  The most noticeable difference includes the ability to use a custom classifier for the recognition, as well as a change from the Classifier object to the Score object.

Please feel free to test and pull.  Cheers,
Joseph


germanattanasio:
Hi @josephlam1984, thanks for the PR but we need to keep v1 and v2. You are proposing to replace v1 with v2.

There is also another problem which is that you are doing the PR to `master` instead of `dev` :disappointed: 

I'll close this and add support for v2 soon.
If you still want to work on this please read the `CONTRIBUTING.md`.

Thanks again for your work.

PR #154, opened 2016-01-10
[tradeoff-analytics] added missing fields and missing serialized names

amiddavid:
option - added a serialized name for decriptionHtml as description_html
solution - fixed names and added missing fields

pls note, I've tried maven test on this, but no tests were picked up and I couldnt find any...


germanattanasio:
@amiddavid, The changes are good but the pull request should be to `dev` not `master`

amiddavid:
should I create a new one for dev? or is there a way to transfer the request?

germanattanasio:
Try with a new one. `dev` is different to `master` so you may need to sync your local `master` with the remote `dev` so that the new PR only reflects your chances

PR #155, opened 2016-01-10
Tradeoff Analytics - fixed serialization names + missing fields

amiddavid:
solution - fixed names and added missing fields
option - added a serialized name for decriptionHtml

### DIFF
--- a/src/main/java/com/ibm/watson/developer_cloud/tradeoff_analytics/v1/model/Option.java
+++ b/src/main/java/com/ibm/watson/developer_cloud/tradeoff_analytics/v1/model/Option.java
@@ -19,6 +19,7 @@ package com.ibm.watson.developer_cloud.tradeoff_analytics.v1.model;
 import java.util.HashMap;
 
 import com.google.gson.GsonBuilder;
+import com.google.gson.annotations.SerializedName;
 
 
 public class Option {
@@ -26,6 +27,7 @@ public class Option {
 	private String key;
 	private String name;
 	private HashMap<String,String> values;
+	@SerializedName("description_html")
 	private String descriptionHtml;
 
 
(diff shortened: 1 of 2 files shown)


amiddavid:
I hope I did it correctly. 
I am new to git, so if I need to do the merges locally,
I would appreciate if you could share the command line commands.
Something along
changing my clone origin from the master to dev, as I initially cloned the master

germanattanasio:
I've merged the changes locally but looks like the test are failing. Seems like you are tring to parse `status_cause` as String but the API returns
```json
"status_cause" : {
  "message" : "A column of a option is out of range. Option \"3\" has a value in column \"rDate\" which is:\"2020-05-29T00:00:00Z\" while the column range\" is: [Thu Apr 29 00:00:00 UTC 2010,Sat Apr 29 00:00:00 UTC 2017]",
  "error_code" : "RANGE_MISMATCH",
  "tokens" : [ "rDate", "2020-05-29T00:00:00Z", "[Thu Apr 29 00:00:00 UTC 2010,Sat Apr 29 00:00:00 UTC 2017]" ]
}
```

I'm going to take a look at the POJOs that we use for Tradeoff Analytics and add one for `status_cause`.

What do you think?

amiddavid:
My fault. Totally forgot about it being an object.
Since its my mistake
I will try to get my local version synced against dev and then add the POJO and run the tests.

You don't need to clean my mess.

germanattanasio:
Don't worry I already fixed. It was just 2 lines of code because the POJO was there just that it wasn't being used.

germanattanasio:
I'm going to work on the Tradeoff ANalytics javadoc, I'm trying to add examples to it like the Gson javadoc: https://google-gson.googlecode.com/svn/trunk/gson/docs/javadocs/com/google/gson/Gson.html

I'm probably going to do a release later today with all the changes

PR #183, opened 2016-02-09
Option for "Return Type" to be JSON in Relationship Extraction

tanmayb123:
Some developers (like me) prefer JSON over XML, and the relationship extraction API does not allow users to choose their preferred return type.

### DIFF
--- a/src/main/java/com/ibm/watson/developer_cloud/relationship_extraction/v1/RelationshipExtraction.java
+++ b/src/main/java/com/ibm/watson/developer_cloud/relationship_extraction/v1/RelationshipExtraction.java
@@ -38,6 +38,9 @@ public class RelationshipExtraction extends WatsonService {
 
   /** The dataset. */
   private Dataset dataset;
+  
+  /** The return type. */
+  private String rt = "xml";
 
   /**
    * Instantiates a new relationship extraction service.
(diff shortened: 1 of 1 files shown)


germanattanasio:
The pull request looks good but can you do it to `dev` instead of `master`. We have the latest code in `dev`.

src/main/java/com/ibm/watson/developer_cloud/relationship_extraction/v1/RelationshipExtraction.java
```diff
@@ -96,5 +99,15 @@ public Dataset getDataset() {
...
     this.dataset = dataset;
   }
+  
+  /**
+   * Sets the return type.
+   * 
+   * @param rt the new rt
+   */
+  
+  public void setReturntype(final String rt) {
```
germanattanasio: `setReturnType` not `setReturntype` and make sure you use `returnType` instead of `rt`.

Can you clarify the different returns types in the javadoc or maybe create an `Enum` class with them ?

tanmayb123: Done. Will create a new pull request.

PR #185, opened 2016-02-09
Option for "Return Type" to be JSON in Relationship Extraction

tanmayb123:
Some developers (like me) prefer JSON over XML, and the relationship extraction API does not allow users to choose their preferred return type.


src/main/java/com/ibm/watson/developer_cloud/relationship_extraction/v1/RelationshipExtraction.java
```diff
@@ -96,5 +99,18 @@ public Dataset getDataset() {
   public void setDataset(final Dataset dataset) {
     this.dataset = dataset;
   }
+  
+  /**
+   * Sets the returnType.
+   * 
+   * @param returnType the new returnType
+   */
+  public void setReturnType(final REReturnType returnType) {
```
germanattanasio: Please don't use acronyms. `ReturnType` is fine.

germanattanasio:
This is looking really good!

tanmayb123:
Thanks! And I've removed the RE acronym from REReturnType.

germanattanasio:
Thanks!

PR #208, opened 2016-03-07
remove prettyprint  as it cause serious network latency no large inputs

amiddavid:


germanattanasio:
closing this since I will be doing the work on #209

Issue #200, opened 2016-02-15
[alchemy-vision] Add support for text extraction


germanattanasio:
POST /image/ImageGetRankedImageSceneText Identifies text in the POSTed image
GET /url/UrlGetRankedImageSceneText Identifies text in the URL

The documentation is: http://www.ibm.com/smarterplanet/us/en/ibmwatson/developercloud/alchemyvision/api/v1/#

mishrasambit:
I tried this extending `com.ibm.watson.developer_cloud.service.WatsonService.`
and using the following code snippet, but I am getting error response
SEVERE: https://gateway-a.watsonplatform.net/calls/image/ImageGetRankedImageSceneText?outputMode=json&imagePostMode=raw, status: 400, error:

```
public <T extends AlchemyGenericModel> T imageGetRankedImageSceneText(
			File image, Class<T> returnType) {
		final Map<String, Object> params = new HashMap<String, Object>();
		params.put("imagePostMode", "raw");
		params.put("outputMode", "json");
		
		final String path = "/image/ImageGetRankedImageSceneText";
		
		final RequestBuilder requestBuilder = RequestBuilder.post(path);
		requestBuilder.withBody(RequestBody.create(HttpMediaType.BINARY_FILE,image));
		for (final String param : params.keySet()) {
			requestBuilder.withQuery(param, params.get(param));
		}
        return executeRequest(requestBuilder.build(), returnType);
}
```

germanattanasio:
You should try to update `AlchemyLanguage.java` not `WatsonService` and it should be pretty straitforward. 

In any case I'm planning on working on this this week.

PR #215, opened 2016-03-16
[alchemy-vision] adding support for scene text issue #200

l2fprod:
new getImageSceneText(file | url) on AlchemyVision to call the new http://www.ibm.com/smarterplanet/us/en/ibmwatson/developercloud/alchemyvision/api/v1/?curl#imagegetrankedimagescenetext

updated unit test with the test image from the AlchemyVision demo http://vision.alchemy.ai/


germanattanasio:
Thanks @l2fprod !

PR #222, opened 2016-03-29
automatically create latest/ symlink in docs for tagged builds

nfriedly:
### Summary

This is to answer the docs team's request to be able to link to the latest release's javadoc without continually having to update things. It should also make life easier on us.

### Other Information

This actually points to the latest tag, so if you, say, release v3.0.0 and then go back and publish v2.9.1 with a bugfix, the latest/ link will change to v2.9.1 until you release a new v3.x. (Or you change it manually in the `gh-pages` branch.)

I'm sending this as a PR rather than committing it so that @germanattanasio can review it and choose when to test it on a new release.


nfriedly:
@germanattanasio can you do a version bump to validate this? Also, I see a [`latest-docs-symlink`](http://watson-developer-cloud.github.io/java-sdk/docs/latest-docs-symlink/) folder which is actually for this branch... we should delete that.

PR #223, opened 2016-03-30
New AuthenticationStrategy

stevenoh93:
### Summary

Allow custom Authentication strategy (i.e. custom method to provide token)

### Other Information

By default, BasicAuthentication is used. No tests have been done with custom auth strategies

### DIFF
--- a/src/main/java/com/ibm/watson/developer_cloud/service/AuthenticationStrategy.java
+++ b/src/main/java/com/ibm/watson/developer_cloud/service/AuthenticationStrategy.java
@@ -1,0 +1,30 @@
+package com.ibm.watson.developer_cloud.service;
+
+public abstract class AuthenticationStrategy {
+	private String token;
+	private String apiKey;
+	private String serviceName;
+	
+	public void setApiKey(String a) {
+		apiKey = a;
+	}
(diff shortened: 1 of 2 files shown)


germanattanasio:
@stevenoh93 can you write here the code that developers will write when using OAuth? and Basic Auth?
You can assume that the Strategy is implemented already so that you only write the part that uses the Java-sdk.

stevenoh93:
By default, the service would use Basic Auth:
```java
SpeechToText service = new SpeechToText();
service.setUsernameAndPassword("<username>", "<password>");
```
If you want to use some other custom Auth startegy, update the AuthStartegy:
```java
// CustomStrategy extends AuthenticationStrategy
CustomStrategy myStrategy = new CustomStrategy();
SpeechToText service = new SpeechToText();
service.setAuthenticationStrategy(myStrategy);
```

germanattanasio:
I've been looking at this and I don't think that it will actually add support for other types of authentications. At the end seems like you always send an API key using the `Authorization` header.

stevenoh93:
Yes, the point was to provide different methods to obtain the API key. Sorry for the late response

Issue #235, opened 2016-04-11
[AlchemyVision] ImageKeyword does not expose knowledge graph data


jasoma:
The [AlchemyVision#getImageKeywords](https://github.com/watson-developer-cloud/java-sdk/blob/master/src/main/java/com/ibm/watson/developer_cloud/alchemy/v1/AlchemyVision.java#L200) has an optional parameter for including knowledge graph data in the response.

But the [ImageKeyword](https://github.com/watson-developer-cloud/java-sdk/blob/master/src/main/java/com/ibm/watson/developer_cloud/alchemy/v1/model/ImageKeyword.java) model objects returned do not have accessors for this information anywhere making it impossible to recover this data using the api client.

PR #236, opened 2016-04-11
add a hierarchy property and modify the adapter to read it #235

jasoma:
### Summary

Adds the `hierarchy` field and accessors to the `ImageKeyword` class so this data from the api can be exposed. See https://github.com/watson-developer-cloud/java-sdk/issues/235

### Other Information

Did not do a full test since I do not have values for the entire `config.properties` list. I ran only the modified test using `gradle test -Dtest.single=AlchemyVisionIT` and my alchemy api key and verified it passed.


germanattanasio:
Great work @jasoma, thanks !

PR #241, opened 2016-04-13
Language Translation - Allow users to use a "Language" Enum instead of "String".

### LINKED ISSUE #225: Add constants for languages in Language Translation
The `LanguageTranslation` service API in the Java SDK does not define any constants for supported languages. For example, right now you have to write:

``` java
service.translate("hello", "en", "es");
```

Rather, having a set of pre-defined constants (an enum most likely) of languages supported by the service would be more helpful:

``` java
service.translate("hello", Language.EN, Language.ES);
```

tanmayb123:
Allow users to use an Enum, instead of Strings (i.e. "Language.ENGLISH" instead of  "en")
This should fix #225

### DIFF
--- a/src/main/java/com/ibm/watson/developer_cloud/language_translation/v2/LanguageTranslation.java
+++ b/src/main/java/com/ibm/watson/developer_cloud/language_translation/v2/LanguageTranslation.java
@@ -253,29 +253,7 @@ public class LanguageTranslation extends WatsonService {
   public TranslationResult translate(final String text, final Language source, final Language target) {
     Validate.isTrue(source != null, "source cannot be null");
     Validate.isTrue(target != null, "target cannot be null");
-    return translateRequest(text, null, langToStr(source), langToStr(target));
-  }
-
-  /**
-   * Convert "Language" enum to a String.
-   *
-   * @param lang The language enum to convert to String.
-   */
-  private String langToStr(Language lang) {
-    if (lang == Language.ARABIC) {
-      return "ar";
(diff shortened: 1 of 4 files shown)


tanmayb123:
@GrapeBaBa Done. Thanks!

GrapeBaBa:
@tanmayb123 That is not I mean. I paste the code, if/else always bad smell.

```
package com.ibm.watson.developer_cloud.language_translation.v2.model;

public enum Language {

    ARABIC {
        @Override
        public String toString() {
            return "ar";
        }
    }, ENGLISH {
        @Override
        public String toString() {
            return "en";
        }
    }, SPANISH {
        @Override
        public String toString() {
            return "es";
        }
    }, FRENCH {
        @Override
        public String toString() {
            return "fr";
        }
    }, ITALIAN {
        @Override
        public String toString() {
            return "it";
        }
    }, PORTUGUESE {
        @Override
        public String toString() {
            return "pt";
        }
    };

    @Override
    public String toString() {
        return "";
    }
}
```

tanmayb123:
Thanks, done.

tanmayb123:
Also, should we have things like "Language.ENGLISH" or "Language.EN"?

germanattanasio:
Language.ENGLISH

Rule #1: NEVER EVER EVER use acronyms.
:P

tanmayb123:
Got it!

germanattanasio:
I'm merging the changes now

Thanks @tanmayb123

PR #269, opened 2016-05-05
Add Bluemix test headers when running tests

gjs29:


germanattanasio:
beautiful as usual  @gjs29

PR #274, opened 2016-05-10
Updated Speech to Text for consistency

relisher:
### Summary

Speech to text instructions inconsistent with API Reference. While previous version would work for wav, that would not work for raw or ogg data, as further specification is necessary to use those file formats, which is incompatible with the information with the Github guide. This update is now consistent with the API Reference

### Other Information

Thank you for the help with IBM STT at TechCrunch Disrupt NY

### DIFF
--- a/README.md
+++ b/README.md
@@ -382,7 +382,7 @@ File audio = new File("src/test/resources/sample1.wav");
 RecognizeOptions options = new RecognizeOptions.Builder()
   .continuous(true)
   .interimResults(true)
-  .contentType(HttpMediaType.AUDIO_WAV)
+  .contentType(audio/wav)
   .build();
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -382,7 +382,7 @@ File audio = new File("src/test/resources/sample1.wav");
 RecognizeOptions options = new RecognizeOptions.Builder()
   .continuous(true)
   .interimResults(true)
-  .contentType(HttpMediaType.AUDIO_WAV)
+  .contentType(audio/wav)
```
germanattanasio: This won't compile.

germanattanasio:
I think the example works because it's using `WAV`. I've fixed an issue related to the use of `rate` when the it's raw audio.
See: 3a11f3fb71efc08c516362d81feb64210099e2ef

PR #290, opened 2016-05-24
Use a proper format for the User-Agent

colindean:
Per an internal discussion, change the User-Agent to be in the valid format for a User-Agent.


colindean:
I may be a little ahead of the game with the name change, @germanattanasio.

germanattanasio:
what if we use `watson-apis` is that against the rules?

colindean:
As in, `watson-apis-java-sdk/3.0.0-RC1`? That's valid.

germanattanasio:
Yeah I would prefer that one.

colindean:
Okay, I'll make that change tomorrow.

PR #293, opened 2016-05-25
Improve Text-To-Speech and Speech-To-Text tests

max-vogler:
### Summary

As discussed with @germanattanasio, I migrated tests to OkHttp's MockWebServer and did some general refactoring.


germanattanasio:
@thanks max!

PR #295, opened 2016-05-26
[text-speech] add getVoice() #287

ndrlslz:
### Summary
add getVoice(String voiceName) function


germanattanasio:
thanks @ndrlslz !

Issue #221, opened 2016-03-28
[text-to-speech] Add support for customizations


germanattanasio:
Add support for customizations. See: https://watson-api-explorer.mybluemix.net/apis/text-to-speech-v1
There are new endpoints like `/v1/pronunciation` and `/v1/customizations`

germanattanasio:
fixed by @max-vogler

PR #297, opened 2016-05-27
[text-to-speech] Added custom voice models & custom translations

max-vogler:
### Summary

Added support for TextToSpeech customizations (custom voice models and word pronunciations)

See: #221

### Todo
- [x] Documentation
- [x] Unit tests
- [x] Convert Long timestamps in CustomVoiceModel to Java Dates


src/main/java/com/ibm/watson/developer_cloud/text_to_speech/v1/model/CustomVoiceModel.java
```diff
@@ -0,0 +1,151 @@
...
+
+  private String name;
+
+  private String description;
+
+  private String language;
+
+  private String owner;
+
+  private Timestamp created;
```
germanattanasio: Let's use `Date` instead of `Timestamp`. I think `GsonSingleton` can handle timestamps

src/main/java/com/ibm/watson/developer_cloud/text_to_speech/v1/Customizations.java
```diff
@@ -0,0 +1,226 @@
...
+ * <p>Use {@code new TextToSpeech().getCustomizations()} to instantiate this
+ * service.</p>
+ * 
+ * @version v1
+ * @see <a href=
+ *      "http://www.ibm.com/smarterplanet/us/en/ibmwatson/developercloud/doc/text-to-speech/custom-intro.shtml">
+ *      Customizations</a>
+ *
+ */
+public class Customizations {
```
germanattanasio: You are adding a Customizations class that can do API calls and that's a problem because we only do API calls from the service class (`TextToSpeech`). This is a design decision that doesn't follow OOP practices but it's required because we charge users when they do an API Call so we want to make sure that the user knows when a method could incur a charge.
Imagine a scenario when we use OOP and do a:

```java
voice.getName()
```

How do you know without looking at the `src` that we will not charge you and translate that method to a `/v1/voices/{id}`?

Let's move the customization calls to `TextToSpeech`

src/main/java/com/ibm/watson/developer_cloud/text_to_speech/v1/model/CustomVoiceModel.java
```diff
@@ -0,0 +1,151 @@
...
+
+/**
+ * A customized voice model, that allows users to specify custom pronunciations
+ * for Waton's Text to Speech API.
+ * 
+ * @see <a href=
+ *      "http://www.ibm.com/smarterplanet/us/en/ibmwatson/developercloud/doc/text-to-speech/custom-intro.shtml">
+ *      Customization</a>
+ */
+public class CustomVoiceModel extends GenericModel {
```
germanattanasio: This class should probably extend `Voice`.
What do you think @max-vogler ?

src/main/java/com/ibm/watson/developer_cloud/text_to_speech/v1/Customizations.java
```diff
@@ -0,0 +1,226 @@
...
+  /**
+   * Saves or updates multiple custom word translations.
+   * 
+   * @param model
+   *          the CustomVoiceModel
+   * @param translations
+   *          the translations to be saved or updated
+   * @return
+   */
+  public ServiceCall<Void> saveWords(CustomVoiceModel model, List<CustomTranslation> translations) {
```
germanattanasio: You can combine this method with `saveWord(CustomVoiceModel model, CustomTranslation translation)` using 3 dots in the method signature.

`saveWord(CustomVoiceModel model, CustomTranslation ...translation)`

users will be able to call it with one or more `CustomTranslation`s

Internally you handle it as an array

src/test/java/com/ibm/watson/developer_cloud/text_to_speech/v1/CustomizationsIT.java
```diff
@@ -0,0 +1,212 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.fail;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.List;
+
+import org.apache.commons.io.IOUtils;
```
germanattanasio: I don't think we have `org.apache.commons.io.IOUtils` in our classpath. I was looking at the `pom.xml` and we only have `lang3`. See https://github.com/watson-developer-cloud/java-sdk/blob/master/pom.xml#L46-50

germanattanasio:
@max-vogler this is a really good job! I left some comments. Let me know if you have questions!

src/test/java/com/ibm/watson/developer_cloud/text_to_speech/v1/CustomizationsIT.java
max-vogler: It's a transitive dependency (`org.mock-server:mockserver-core` > `commons-io`). Since we're stopping to use org.mock-server, I implemented the method on my own.

src/main/java/com/ibm/watson/developer_cloud/text_to_speech/v1/model/CustomVoiceModel.java
max-vogler: I had the same idea, but ended up with a different design: Even though they are named similar, `CustomVoiceModels` do not inherit any property of `Voice` in the OOP sense (see [TTS documentation](http://www.ibm.com/smarterplanet/us/en/ibmwatson/developercloud/doc/text-to-speech/custom-intro.shtml)). I'm not 100% happy with the naming either, so I just followed the official documentation. The `Voice` class is what you'd expect, while a `CustomVoiceModel` is a dictionary of words and their pronunciations. So a `CustomVoiceModel` is not a specialized form of a `Voice`, but can be used as plugin to modify the speech output of different Voices.

src/main/java/com/ibm/watson/developer_cloud/text_to_speech/v1/Customizations.java
max-vogler: Thanks for the great explanation! Done.

src/main/java/com/ibm/watson/developer_cloud/text_to_speech/v1/model/CustomVoiceModel.java
max-vogler: I considered this as well. If you give me further input, I can change it. The customizations API diverges from the other APIs by using unix timestamps instead of a date strings. So we could:

1. Use `Date` objects and adapt `DateDeserializer` to parse Unix timestamps: Downside: When serializing a CustomVoiceModel using GSON, the `DateSerializer` displays the Dates as date string and not as timestamp
2. Use `Date` objects and create a custom TypeAdapter for CustomVoiceModel only
3. Use `Timestamps`, which is a thin-layer on `Date`. This allows us to use an existing TypeAdapter. The methods `getCreated()` and `getLastModified()` still return Date objects, so the API user does not really see a difference.

What do you suggest?

src/main/java/com/ibm/watson/developer_cloud/text_to_speech/v1/model/CustomVoiceModel.java
germanattanasio: make sense

src/main/java/com/ibm/watson/developer_cloud/text_to_speech/v1/model/CustomVoiceModel.java
germanattanasio: What if we use `Long` for the timestamps. I don't like the idea of using a `java.sql` class but I think that having `Date`s will complicate things

Issue #298, opened 2016-05-29
Proposed to increase the new construction method


handong890:
Some time we using spring for our project and we want  control them keep one instance.
**So please add all constructor method with username and password**

For example com.ibm.watson.developer_cloud.dialog.v1.DialogService
` public DialogService(String username,String password) {
		    super("dialog");
		    setEndPoint(URL);
		    setUsernameAndPassword(username, password);
		  }`

if have this then I can config it to our spring project
`<bean id="dialogService" class="com.ibm.watson.developer_cloud.dialog.v1.DialogService">
		 <constructor-arg name="username" value="${dialogusername}" />
        <constructor-arg name="password" value="${dialogpassword}"  />
	</bean>	`

germanattanasio:
fixed by @ndrlslz

PR #300, opened 2016-05-30
add constructor method with username and password. #298

ndrlslz:
### Summary
add constructor method with username and password. #298

### DIFF
--- a/src/main/java/com/ibm/watson/developer_cloud/visual_recognition/v2_beta/VisualRecognition.java
+++ b/src/main/java/com/ibm/watson/developer_cloud/visual_recognition/v2_beta/VisualRecognition.java
@@ -45,7 +45,8 @@ import okhttp3.RequestBody;
  * can organize image libraries, understand an individual image, and create custom classifiers for
  * specific results that are tailored to your needs.
  * 
- * @version v2
+ * @version v2_beta
+ * @version_data 2015-12-02
  * @see <a href=
  *      "http://www.ibm.com/smarterplanet/us/en/ibmwatson/developercloud/visual-recognition.html">
  *      Visual Recognition</a>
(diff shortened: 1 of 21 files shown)


src/main/java/com/ibm/watson/developer_cloud/visual_recognition/v3/VisualRecognition.java
```diff
@@ -90,6 +90,19 @@ public VisualRecognition(String versionDate) {
   }
 
   /**
+   * Instantiates a new Visual Recognition V2 service by username and password.
```
germanattanasio: `V2` should be `v3`

src/main/java/com/ibm/watson/developer_cloud/visual_recognition/v2_beta/VisualRecognition.java
```diff
@@ -82,6 +82,18 @@ public VisualRecognition(String versionDate) {
     this.versionDate = versionDate;
   }
 
+  /**
+   * Instantiates a new Visual Recognition V2 service by username and password.
```
germanattanasio: `v2` should be `v2-beta`

germanattanasio:
I'm happy with merging this once we have the two issues mentioned above fixed.

Can you also add a `@see <link to service page>` and `@version <api version>` on each contructor?

For Tone Analyzer `v3-beta` it will be: 
```
/**
 * The IBM Watson Tone Analyzer service uses linguistic analysis to detect emotional tones, social
 * propensities, and writing styles in written communication. Then it offers suggestions to help the
 * writer improve their intended language tones.
 * 
 * @version v3-beta
 * @see <a href=
 *      "http://www.ibm.com/smarterplanet/us/en/ibmwatson/developercloud/tone-analyzer.html"> Tone
 *      Analyzer</a>
 */
```
you can also list the `version_date` so that users know what value they can use.

In Tone Analyzer v3 it will be `2016-05-19`

ndrlslz:
@germanattanasio i did not catch your point about adding `@see` and `@version` on each constructor.

Because i can find theses tags in the beginning of class.

Is it necessary to add theses on the constructor repeatedly?

germanattanasio:
good point @ndrlslz... I forgot about those tags being in the class...

PR #302, opened 2016-05-31
add example of categorical column

boazdavid:
I added an example of how to use the sdk for Tradeoff Analytics with a caegorical column.
according to the DW question https://developer.ibm.com/answers/questions/275792/how-to-use-categoricalcolumn-is-there-any-example.html


boazdavid:
by mistake, the 2 commits went into the same PR.
please note that b7ec97b is a bug fix.
When the fix will be available in maven repo?

germanattanasio:
@boazdavid I'm trying to do a release either today or tomorrow.

btw, there is an issue related to how we serialize/deserialize columns for TA.
See #155, can you take a look?

Issue #322, opened 2016-06-08
Potential Travis and Codecov issues


max-vogler:
**There are two potential issue with Travis and Codecov:**

- [x] The code coverage displayed on Codecov seems to exclude unit test coverage (see [CredentialUtils.java](https://codecov.io/gh/watson-developer-cloud/java-sdk/src/d4848b3652a0bd00f29676478c4dd317047eb252/src/main/java/com/ibm/watson/developer_cloud/util/CredentialUtils.java) vs [CredentialUtilsTest.java](https://github.com/watson-developer-cloud/java-sdk/blob/master/src/test/java/com/ibm/watson/developer_cloud/util/CredentialUtilsTest.java)
- [x] The Travis build seems to pass, even though the integration tests fail (see [Travis build log](https://travis-ci.org/watson-developer-cloud/java-sdk/jobs/136002119#L716))

max-vogler:
Codecov still does not calculate the correct sum of UT and IT coverage. For demonstration purposes I set up a simple [test branch](https://github.com/watson-developer-cloud/java-sdk/tree/test), which contains a [`Pojo` class](https://github.com/watson-developer-cloud/java-sdk/blob/test/src/main/java/com/ibm/watson/developer_cloud/Pojo.java), and two tests [`UnitTest`](https://github.com/watson-developer-cloud/java-sdk/blob/test/src/test/java/com/ibm/watson/developer_cloud/UnitTest.java) and [`IntegrationIT`](https://github.com/watson-developer-cloud/java-sdk/blob/test/src/test/java/com/ibm/watson/developer_cloud/IntegrationIT.java). The Pojo class should be covered 100% by both tests, but Codecov [displays 78%](https://codecov.io/gh/watson-developer-cloud/java-sdk/src/test/src/main/java/com/ibm/watson/developer_cloud/Pojo.java). This is clearly a Codecov issue, because [Cobertura reports 100% on my local machine](https://cloud.githubusercontent.com/assets/864168/15935935/72930430-2e69-11e6-9fe9-38c97a42f8b7.png).

max-vogler:
I opened a support ticket: https://github.com/codecov/support/issues/227

PR #323, opened 2016-06-08
Enforce travis.sh failure when maven fails

max-vogler:
### Summary

#322: The Travis build seems to pass, even though the integration tests fail (see [Travis build log](https://travis-ci.org/watson-developer-cloud/java-sdk/jobs/136002119#L716))

### DIFF
--- a/.utility/travis.sh
+++ b/.utility/travis.sh
@@ -1,10 +1,12 @@
 #!/bin/bash
 
+set -e # enforces the script to fail as soon as one command fails
+
 if [ "${TRAVIS_PULL_REQUEST}" = "false" ]; then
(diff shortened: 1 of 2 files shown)


max-vogler:
@germanattanasio: If you agree that failing integration tests should make the build fail, you can merge this even with unsuccessful checks (they fail because of #324)

germanattanasio:
will the call to cobertura return a status code != 0 ? because when we do a release we don't run the integration tests. The release process doesn't have credentials so we should make sure that a fresh checkout doesn't fail when running ```mvn test```

max-vogler:
The behavior of `mvn test` has not been changed. Before this fix, `mvn verify` returned `0` even with integration test errors. Now `mvn verify` returns `1` when integration tests fail.

Issue #316, opened 2016-06-07
Migrate to OkHttp's MockWebServer


max-vogler:
# Summary

Remove `MockServer` and use `MockWebServer` in unit tests

- [x] `natural_language_classifier`
- [x] `personality_insights`
- [x] `service.GenericServiceTest`
- [x] `tone_analyzer`

## Backlog:
- [x] `document_conversion`
- [x] Remove dependency, clean up WatsonServiceUnitTest

germanattanasio:
I'm moving this to the backlog. The Document conversion team will probably do this

PR #328, opened 2016-06-08
[WIP] Improve integration test execution

max-vogler:
### Summary

This pull requests includes unit tests in the code coverage in a simple way. The downside is that unit tests are now effectively executed twice for the `mvn cobertura:cobertura-integration-tests` goal.

Removing the MockServer in #316 allows a class-based parallelization of the integration tests. This means that all methods inside one class are executed sequentially, but multiple test classes may be executed in parallel. I had promising results with a parallelization factor of **8** on my local machine, cutting down integration test time to 1:07 min

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -171,6 +171,10 @@
 				<artifactId>maven-failsafe-plugin</artifactId>
 				<version>2.19</version>
 				<configuration>
+					<!--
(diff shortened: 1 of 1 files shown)


germanattanasio:
I added this change on another commit

PR #349, opened 2016-06-14
Fix javax android compatability

Hsaylor:
This pull request fixes the issues developers targeting older Android versions were seeing due to the use of `javax`. This has been tested with API 19 (Android 4.4), API 21 (Android 5.0), and API 23 (Android 6.0).


germanattanasio:
beautiful @Hsaylor

PR #352, opened 2016-06-15
Update R&R SolrJ example

gjs29:
Use version 5.5.1 of SolrJ
Execute calls to the config API synchronously


max-vogler:
LGTM :+1:

PR #354, opened 2016-06-17
[text-to-speech] add customization example

ndrlslz:
### Summary

regarding TextToSpeech, we have supported custom voice model.
So add the example about this.


max-vogler:
Tested it, LGMT :+1: 

@ndrlslz: If you find some spare time, you could replace `new TextToSpeech()` with `new TextToSpeech(username, password)` and add speech synthesis with the custom voice model to the example.

PR #359, opened 2016-06-21
Remove Apache Commons Lang dependency

max-vogler:
### Summary

Following the efforts to keep the method count and Android binary size low, it is now safe to remove Apache Commons Lang as dependency. This is a reduction of 3,582 methods and 435 kb :tada:  

To keep a pleasant testing experience, Apache Commons Lang is still included in the tests.


germanattanasio:
niceeee

PR #366, opened 2016-06-27
Add Solr resize API to R&R

anathashavit:
Add a cluster resize API that allows posting a resize operation and getting information on a cluster's size using the Retrieve&Rank service.  
In addition there is an example file that shows how to use the new API. 

### DIFF
--- a/src/main/java/com/ibm/watson/developer_cloud/retrieve_and_rank/v1/model/SolrClusterResizeRequest.java
+++ b/src/main/java/com/ibm/watson/developer_cloud/retrieve_and_rank/v1/model/SolrClusterResizeRequest.java
@@ -1,0 +1,42 @@
+/*
+ * BEGIN_COPYRIGHT
+ *
+ * IBM Confidential OCO Source Materials
+ *
+ * 5725-Y07 (C) Copyright IBM Corp. 2016 All Rights Reserved.
+ *
+ * The source code for this program is not published or otherwise divested of its trade secrets,
+ * irrespective of what has been deposited with the U.S. Copyright Office.
+ *
+ * END_COPYRIGHT
+ */
+package com.ibm.watson.developer_cloud.retrieve_and_rank.v1.model;
+
+
+import static com.ibm.watson.developer_cloud.retrieve_and_rank.v1.model.ApiConstants.*;
+
+import com.google.gson.annotations.SerializedName;
+
+/**
+ * A value type for the JSON body provided when creating a Solr resize request.
+ */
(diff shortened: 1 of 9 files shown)


src/main/java/com/ibm/watson/developer_cloud/retrieve_and_rank/v1/model/SolrClusterResizeRequest.java
```diff
@@ -0,0 +1,42 @@
+/*
+ * BEGIN_COPYRIGHT
```
germanattanasio: The headers should make the ones we use in the project

germanattanasio:
Can we add unit tests to the new functionality ?

src/main/java/com/ibm/watson/developer_cloud/retrieve_and_rank/v1/model/SolrClusterResizeRequest.java
anathashavit: Done

germanattanasio:
@anathashavit the Retrieve and Rank integration test will take a lot of time. Can you add a unit test to make sure the request match the API specs? You can take a look at the other unit tests if you need.

anathashavit:
@germanattanasio looking at the R&R test package I see that all other API's are testes the same way. 
There are a couple of unit tests for utility classes used by the API but not foe the API itself since it would be very difficult to verify the API using only a unit test.

germanattanasio:
What I'm asking is something like:

```java
    // mock api call with a default json response
    ToneAnalysis mockResponse = loadFixture(FIXTURE, ToneAnalysis.class);
    server.enqueue(jsonResponse(mockResponse));
    
    // execute request
    ToneAnalysis serviceResponse = service.getTone(text, null).execute();

    RecordedRequest request = server.takeRequest();

    assertEquals(TONE_PATH, request.getPath());
    assertNotNull(request.getHeader(HttpHeaders.AUTHORIZATION));
    assertEquals(serviceResponse, mockResponse);
    assertEquals(HttpMediaType.APPLICATION_JSON, request.getHeader(HttpHeaders.ACCEPT));
```
Take a look at the unit tests for [Tone Analyzer](https://github.com/watson-developer-cloud/java-sdk/blob/master/src/test/java/com/ibm/watson/developer_cloud/tone_analyzer/v3/ToneAnalyzerTest.java)

The APIs that you are wrapping in `resizeSolrCluster` and `getSolrClusterResizeStatus` should be tested with a unit test.

Issue #367, opened 2016-07-01
TypedRelation service call in Alchemy Language Java SDK not returning entities


nastacio:
I am using the [AlchemyLanguage] service, more specifically the Typed Relations service.

I am using version 3.0.2-SNAPSHOT

        <repository>
            <id>sonatype</id>
            <url>https://oss.sonatype.org/content/repositories/snapshots</url>
        </repository>

        <dependency>
            <groupId>com.ibm.watson.developer_cloud</groupId>
            <artifactId>java-sdk</artifactId>
            <version>3.0.2-SNAPSHOT</version>
        </dependency>

Somehow the resulting Typed Relation objects do not contain the "entities" values, just the relation type:

   ```
 @Test
    public void testX() throws IOException {
        AlchemyLanguage service = new AlchemyLanguage();
        service.setApiKey("...");
        service.setEndPoint("https://gateway-a.watsonplatform.net/calls");

        Path input = Paths.get("./src/test/resources/9563489.txt");
        ByteArrayOutputStream baos = new ByteArrayOutputStream();
        Files.copy(input, baos);

        Map<String, Object> params = new HashMap<String, Object>();
        params.put(AlchemyLanguage.TEXT, baos.toString("UTF-8"));
        params.put(AlchemyLanguage.MODEL_ID, "ie-en-news");

        ServiceCall<TypedRelations> trCall = service.getTypedRelations(params);
        TypedRelations typedRelations = trCall.execute();
        List<TypedRelation> trs = typedRelations.getTypedRelations();
        trs.stream().forEach(tr -> {
            System.out.println(tr.getEntities());
            System.out.println("\t" + tr.getType());
        });
    }
```

And the output is:

```
null
	locatedAt
null
	hasAttribute
null
	partOfMany

```

nastacio:
In case it helps (still sorting out issues with my local git repo, will try and submit a pull request later)

Code fix
[Archive.zip](https://github.com/watson-developer-cloud/java-sdk/files/344615/Archive.zip)

Testcase fix.
[AlchemyLanguageIT.java.zip](https://github.com/watson-developer-cloud/java-sdk/files/344619/AlchemyLanguageIT.java.zip)

germanattanasio:
fixed in #403

PR #370, opened 2016-07-02
 TypedRelation service call in Alchemy Language Java SDK not returning entities

nastacio:
🐛

This is a proposed fix for
#367

It takes the "arguments/argument" indirection into account when deserializing the server payloads into TypedRelation Java objects. Without this fix, the "entities" are not deserialized correctly and TypedRelation.getEntities() always returns a null pointer.

I also changed the type of TypedRelation.entities from Entity to TypedEntity, which seemed the original intention of the design all along.

### DIFF
--- a/src/main/java/com/ibm/watson/developer_cloud/alchemy/v1/model/TypedEntitiesAdapter.java
+++ b/src/main/java/com/ibm/watson/developer_cloud/alchemy/v1/model/TypedEntitiesAdapter.java
@@ -28,6 +28,8 @@ public class TypedEntitiesAdapter extends TypeAdapter<List<TypedEntity>> {
 
     @Override
     public void write(JsonWriter writer, List<TypedEntity> value) throws IOException {
+        // No coding the serialization as there is no code relying on this
+        // type to serialize payloads
     }
 
     @Override
--- a/src/main/java/com/ibm/watson/developer_cloud/alchemy/v1/model/TypedEntitiesAdapter.java
+++ b/src/main/java/com/ibm/watson/developer_cloud/alchemy/v1/model/TypedEntitiesAdapter.java
@@ -1,0 +1,58 @@
+package com.ibm.watson.developer_cloud.alchemy.v1.model;
+
+import com.google.gson.TypeAdapter;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonWriter;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+public class TypedEntitiesAdapter extends TypeAdapter<List<TypedEntity>> {
+
+    @Override
+    public void write(JsonWriter writer, List<TypedEntity> value) throws IOException {
+    }
+
+    @Override
+    public List<TypedEntity> read(JsonReader reader) throws IOException {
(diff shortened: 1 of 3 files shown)


germanattanasio:
I feel like this can be done with POJOs and without TypeAdapters. Is that something you have looked into @nastacio ?

nastacio:
I have not used this Google library before (I know it is popular) , but I understand a plain-POJO 
resolution would require the introduction of new classes to represent the “arguments” and “argument”
elements surrounding the entities. I felt this would only clutter the Java API.

The fundamental problem resides with the payload design itself, which is unnecessarily verbose
in the representation of the entities.

Denilson Nastacio
dnastacio@gmail.com

max-vogler:
@nastacio: I totally agree with your opinion about the payload design - it is complex! That *cluttering*  of the Watson Java SDK is a definitely problem we're facing.

Nevertheless, we used the approach with POJOs throughout the Java SDK and try to keep the amount of custom (de)serializers low. Although unintuitive at first, this is benefitial in this case because of multiple reasons:

 1. Because the Java objects mirror the API responses very closely, we can employ a simple `equals()` check by comparing the JSON representation of Java objects. This requires a `write()` method in `TypeAdapters`, which is currently failing silently in your implementation.

 2. Testing custom (de)serializers is tedious because they virtually consist of branches only. It requires a lot of unit tests to get a good coverage. When not using GSON's default (de)serializer, nasty edge cases have to be handled manually - null handling for example.

 3. The maintainability, reusability, and extensibility of custom (de)serializers is simply bad. Most API changes implicate changes in the Java object anway. Requiring the custom (de)serializers to be changed as well results in duplicate work.

An example of a commit that solved a similar problem is 8a13f41a8d6219246

nastacio:
At some point one must decide what is the fundamental interface for developers: the server payload or the language binding. The design for the server payload goes from verbose  (e.g. the  "arguments/argument" wrappers) to whimsical ( the "part:first | second" elements inside the "argument" element) to inconsistent (the "argument" element only exists in XML, but not in JSON) , so I somehow assumed it would make sense to exorcise some of it in the language binding. 

Points #2 and #3 on maintainability/testing/etc were lost on me. The presumed savings in not writing the tedious test code for null pointers and multiple payload variants were the very cause of this bug. The GSON library glides over misalignments between payload and Java classes and will silently assign null pointers to attributes it cannot find in the payloads, so that you still need to test all of the mapping possibilities even if you did not rely on type adapters.

If the idea is to stick to a literal mapping (thus eliminating the inherent obscurity of type adapters) AND to cut down on testing the payload mapping and variations to the nth degree, then those two purposes may be best served by creating an XSD schema from complete payloads (couple of clicks on a decent IDE) and then using the xjc compiler to automatically generate the POJOs.

I definitely understand the choice of POJOs as a design philosophy that makes the the API mappings more predictable for people familiar with the server payload API, so I mean no criticism towards that approach.

Themandunord:
Hi,

When this request will be merge to a new version of the API ?
I need this entities' information in my project.

Thanks.

germanattanasio:
The API changed since you did this work @nastacio but I managed to update what you did to match the new API and wrote a pull request, see #403

PR #380, opened 2016-07-13
fix for VisualRecognitionExample

enginer:
simple fix for NullPointerException in `com.ibm.watson.developer_cloud.visual_recognition.v3.VisualRecognitionExample` "Classify using the 'Car' classifier" block


kognate:
Can you add some test coverage for this change?  

Thank you.

germanattanasio:
I fixed this by initializing the array in `classifierIds()` instead of the constructor. The VR unit test was also updated.

See #399

PR #384, opened 2016-07-14
:new: Proxy support

MarzlS:
### Summary

Supports the use of a proxy server to access the services.

Sample code for proxy usage:

```
WatsonService.setProxyConfiguration(new ProxyConfiguration("myproxy", 8080));
```
Or with proxy that requires authentication:

```
WatsonService.setProxyConfiguration(new ProxyConfiguration(proxyHost, proxyPort, proxyUser, proxyPassword));
```

### Other Information

Proxy is required for our usage scenario (calling NLC from customer environment)

### DIFF
--- a/src/main/java/com/ibm/watson/developer_cloud/http/ProxyConfiguration.java
+++ b/src/main/java/com/ibm/watson/developer_cloud/http/ProxyConfiguration.java
@@ -1,0 +1,79 @@
+/**
+ * Copyright 2016 IBM Corp. All Rights Reserved.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ * 
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.ibm.watson.developer_cloud.http;
+
+import okhttp3.Credentials;
+
+/**
+ * Stores the proxy configuration in case a proxy needs to be used to access the services.
+ */
+public class ProxyConfiguration {
+
+    private String hostname;
+    private int port;
+    private String credentials;
+    
+    /**
+     * Creates a proxy configuration without authentication.
+     * 
+     * @param hostname     host name of the proxy
+     * @param port        port of the proxy
+     */
+    public ProxyConfiguration(String hostname, int port) {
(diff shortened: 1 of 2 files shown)


germanattanasio:
What's the different between what you did and setting a different service url?

```java
LanguageTranslation service = new LanguageTranslation();
service.setEndPoint("<proxy-url>");
```

MarzlS:
Sorry but I do not understand your question, I want to access the service not the proxy. As far as I know when your client is isolated and needs a proxy to connect to Internet resources you need to configure the proxy on HTTP client level. I am not aware of an URL format that contains both the target URL to connect, the proxy server to use and the proxy authentication credentials to use. But maybe I am missing something.
Or are talking about creating something like a forwarding configuration on the proxy itself?

germanattanasio:
Instead of using a proxy on HTTP level you can change the service url to point to your proxy and make the proxy forward the requests to the real service.

```java
LanguageTranslation service = new LanguageTranslation();
service.setEndPoint("<proxy-url>");

List<Model> models = service.getModels().execute();
```
You can also use basic auth with your proxy since it's provided by the existing implementation.

```java
LanguageTranslation service = new LanguageTranslation();
service.setEndPoint("<proxy-url>");
service.setUsernameAndPassowrd("<proxy-username>", "<proxy-password>");

List<Model> models = service.getModels().execute();
```

The code above will send a request to `<proxy-url>` using basic auth. You only need to make a forwarding rule from your `<proxy-url>` to the real service.

MarzlS:
Ok, thank for clarifying this, I understood now. The problem is that in our case the proxy is maintained by different client IT department and I fear they are rather unflexible here as this requires a change to their production proxy configuration that provides proxy service to all their servers. 

So setting the proxy on HTTP client level in my opinion is the most flexible option (and the proxy configuration on the HTTP library is there for a reason of course). 

Regarding authentication, IMO your suggestion means that proxy user and service user need to have same credentials? An other complication that we would need to request from our clients that they need a proxy user (usually users from their directory service) named same as the service user (may be violating their user naming scheme?) with exactly same credentials...

germanattanasio:
@MarzlS I will merge the changes but I need you to:

- [ ] Improve the JavaDoc
- [ ] Add unit tests that validate the request and that check for null or invalid parameters. We should throw `IllegalArgumentException` if a required parameter is missing.
- [ ] Add a section in the `README.md` describing how to set a `Proxy`
- [ ] Add an example in the `examples` folder 

Finally, I'm worried about extensibility, how does your solution handle cases like when the `host` is an IP address. See square/okhttp#271

vladimir-romaniuk:
Hi.  I also needed proxy support for my customer.
Why this pull request is closed and not merged? Actually it would be good to have proxy support inside java-sdk. Can you give me update how you finished task with proxy support ?

germanattanasio:
@vladimir-romaniuk Please read the README in this project. This code was not merged because the feature was implemented in another Pull request

PR #425, opened 2016-08-10
Conversation v1 Unit tests

Mikemosca:
add new test and a few new negative tests


germanattanasio:
Good job @Mikemosca !

PR #435, opened 2016-08-29
ENH:  Added the customization id for a custom language model

baarongit:
### Summary

I added the functionality to include a customization id so that we can specify a language model to use during the recognize service call.

… call to recognize the audio for speech to text.


germanattanasio:
@baarongit I need you to sign the CLA before I'm able to merge this.

baarongit:
@germanattanasio I have signed the license multiple times but the icon will not update.

Issue #454, opened 2016-10-07
[text-to-speech] Using AudioFormat PCM in method synthesize throws exception accept=audio/l16, status: 406, error: rate parameter missing in mimetype audio/l16


roymitchley:
I am trying to call the text-to-speech synthesize method with AudioFormat.PCM. This is giving me the exception accept=audio/l16, status: 406, error: rate parameter missing in mimetype audio/l16
com.ibm.watson.developer_cloud.service.exception.ForbiddenException: rate parameter missing in mimetype audio/l16

I can not see a way for me to pass the rate parameter as the method parameter for synthesize is a AudioFormat type that is a numeric.

It looks like a similar problem has been solved for speech-to-text service.

https://github.com/watson-developer-cloud/java-sdk/commit/3a11f3fb71efc08c516362d81feb64210099e2ef

I am using version 3.4.0

PR #462, opened 2016-10-14
Add AudioFormat.getPCM() to set the rate

arthurfabre:
### Summary

This allows the normal TextToSpeech API to be used, with PCM
AudioFormats with a rate set.

Deprecate the old PCM constant, as it can't be used safely.

### Other Information

This adds a sane way of using the previous fix to #454 (c78d199ac0cc333354fe753056dbd8789659de7a)


germanattanasio:
Thanks @arthurfabre!

PR #469, opened 2016-10-24
Moved conversation-tone-integration to right folder

vibhasinghal:
### Summary

The tone conversation integration example code was not in appropriate folder. 
Moved it to be right under examples. This makes it's location consistent across all sdks (node, python).


vibhasinghal:
April will submit a pull request.

PR #480, opened 2016-11-17
add new tests for NLC

Mikemosca:
Add new Unit tests for NLC
refactor integration tests

I was not able to modify the . config.properties.enc file.  The classifier id it uses is no longer around.  Created a new one but need to modify that file.  So in that file can you replace it with the following
natural_language_classifier.classifier_id=e82f62x108-nlc-5806


germanattanasio:
Mike can you switch your editor to use 2 spaces indentation. Seems like you are using tabs 😛

PR #512, opened 2017-01-03
Fix environment size enum serialization for Discovery

maniax89:
### Summary

Fixes the environment size enum serialization issue (it was outputting `FREE` instead of `0`) see the API spec -> https://watson-api-explorer.mybluemix.net/apis/discovery-v1#!/Environments/post_v1_environments

### DIFF
--- a/tests/src/test/java/com/ibm/watson/developer_cloud/discovery/v1/DiscoveryServiceIT.java
+++ b/tests/src/test/java/com/ibm/watson/developer_cloud/discovery/v1/DiscoveryServiceIT.java
@@ -132,6 +132,8 @@ public class DiscoveryServiceIT extends WatsonServiceTest {
                     .build();
             CreateEnvironmentResponse createResponse = discovery.createEnvironment(createRequest).execute();
             environmentId = createResponse.getEnvironmentId();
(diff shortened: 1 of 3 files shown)


tests/src/test/java/com/ibm/watson/developer_cloud/discovery/v1/DiscoveryServiceIT.java
```diff
@@ -132,6 +132,8 @@ public static void setupClass() {
                     .build();
             CreateEnvironmentResponse createResponse = discovery.createEnvironment(createRequest).execute();
```
maniax89: this line was failing if you didn't already have an environment created before starting these tests

tests/src/test/java/com/ibm/watson/developer_cloud/discovery/v1/DiscoveryServiceIT.java
```diff
@@ -250,16 +252,10 @@ public void delete_environment_is_successful() {
     }
 
     @Test
-    @Ignore("Environment creation/deletion is expensive")
```
kognate: Shouldn't this one get ignored, too?

PR #556, opened 2017-02-03
fix the gradle check errors

Mikemosca:
fix the gradle check errors


Mikemosca:
I added a line to the readme to run gradle check as a best practice

also fixed an error in eclipse where the alchemy proejct was not imported

Issue #498, opened 2016-12-09
[speech-to-text] Add getCorpus() method


jeffpk62:
The service recently added the equivalent of a getCorpus() method to return information about a single corpus. It returns the same information as getCorpora() but for a single specified corpus only. It's very useful for learning the status of an addCorpus() call, and it's the analog of all of the get one/many methods offered by the service.

jeffpk62:
Done.

PR #558, opened 2017-02-06
Add Speech to Text getCorpus() method

jeffpk62:
Adds the Speech to Text getCorpus() method requested in issue https://github.com/watson-developer-cloud/java-sdk/issues/498.


germanattanasio:
Thanks @jeffpk62

PR #559, opened 2017-02-07
:new:update the conversation API wrapper to refer to the latest version

doconnor78:
Update the Conversation API wrapper classes to include the 2017-02-03 API version


conversation/src/main/java/com/ibm/watson/developer_cloud/conversation/v1/ConversationService.java
```diff
@@ -53,7 +55,7 @@ public ConversationService(final String versionDate) {
     }
 
     Validator.isTrue((versionDate != null) && !versionDate.isEmpty(),
-        "'version cannot be null. Use " + VERSION_DATE_2016_09_20);
+        "'version cannot be null. Use " + VERSION_DATE_2017_02_03);
```
nfriedly: Shouldn't this be `"'version cannot be null. Use ConversationService.VERSION_DATE_2017_02_03");` so that users know to pass in the constant rather than copying the string?

PR #563, opened 2017-02-08
Update gson to v2.8.0

aahlenst:


germanattanasio:
@aahlenst will this update bring new features that we need? I feel like the library is growing in size with features we don't need.

aahlenst:
My motivation was not new features but bug fixes and to ensure that there are no problems with newer versions of gson. I assume that I'm not the only consumer of the Watson Java SDK that's using newer versions of gson (either deliberately or because it's forced upon me by another library). 

I'm fine with skipping the update until there are any incompatibilities that make it impossible to use the Watson Java SDK with a newer version of gson.

Issue #574, opened 2017-02-15
[speech-to-text] Incorrect overwrite parameter sent to add corpus file


jeffpk62:
The `addTextToCustomizationCorpus()` method includes a Boolean `allowOverride` parameter.  Specifying `true` for the parameter passes the query parameter as `allow_override`.  But the query parameter is named `allow_overwrite`, so the service ignores the parameter and still fails on a valid overwrite request.  The name of the parameter and the query string it sends both need to be adjusted.

Issue #573, opened 2017-02-14
[speech-to-text] Add an addCorpus() method


jeffpk62:
The names of the methods for the Java SDK parallel the service functionality and the Node SDK pretty well, with one exception.  Instead of an `addCorpus()` method, the interface offers an `addTextToCustomizationCorpus()` method.  I don't know what happened here, but I'd like to add the simpler `addCorpus()` alternative to go along with `getCorpora()`, `getCorpus()`, and `deleteCorpus()`, as well as `addWords()`, `addWord()`, `getWords()`, `getWord()`, and `deleteWord()`.  I think that it would make the interface look much cleaner.

In the new version of the method, move the optional overwrite parameter to the end of the signature.  The other three parameters are required, so we should list them first.

We should deprecate the longer form of the method once the new version is available.  Of course, the old method and signature will continue to work for as long as the method is provided.

Issue #503, opened 2016-12-12
[speech-to-text] Integration tests are not deleting custom language models


germanattanasio:
From: @nfriedly:
```
hey, there are like a hundred `IEEE-java-sdk-permanent` customizations
for speech to text
I think some code is creating those that shouldn't be
(The node SDK's integration test automatically nukes
anything that doesn't have the word `permanent` in the
name and is > 1 day old before running it's tests - so
the Java SDK may well be leaving behind temporary ones too.) 
```

jeffpk62:
German, I can address this one for you.  Just let me know.

germanattanasio:
Yes please, assign the issues to you and you are set.

jeffpk62:
@germanattanasio, the integration test already had a finally clause that deleted the custom model.  The 140 models that existed were from December and January.  I did the following:

- Updated the name and description of the IT custom model to indicate that it's temporary.  Otherwise, the IT code is already good.
- Verified that the STT tests all pass without the old models and that we're not leaking any new models.
- Deleted 135 of the old custom models.  Five of the models are from early December but are in a locked state that prevents their deletion.  I will follow up with the Speech team.  The five that remain are all safe to delete, and they all generate the following conflict exception in response to attempted deletion:

```
Deleting: js-sdk-test-temporary
Feb 15, 2017 2:42:05 PM com.ibm.watson.developer_cloud.service.WatsonService processServiceCall
SEVERE: DELETE https://stream.watsonplatform.net/speech-to-text/api/v1/customizations/526fe850-
c093-11e6-9657-e53f14d32a55, status: 409, error: Customization '526fe850-c093-11e6-9657-e53f14d32a55'
is currently locked to process your last request.
com.ibm.watson.developer_cloud.service.exception.ConflictException: Customization '526fe850-
c093-11e6-9657-e53f14d32a55' is currently locked to process your last request.
        at com.ibm.watson.developer_cloud.service.WatsonService.processServiceCall(WatsonService.java:426)
        at com.ibm.watson.developer_cloud.service.WatsonService$1.execute(WatsonService.java:180)
        at com.ibm.watson.developer_cloud.speech_to_text.v1.delete.main(delete.java:34)
```

At this point, this can probably be closed, since no further changes are need to the SDK.

jeffpk62:
The Speech team is looking into deleting the locked models.  We can't do anything with them.

germanattanasio:
You have the credentials so I think you can delete them

[...]

PR #575, opened 2017-02-15
Add addCorpus() method and correct allowOverride parameter

jeffpk62:
This PR address two issues for the Speech to Text service:

- It changes the `allowOverride` parameter of the `addTextToCustomizationCorpus()` method to `allowOverwrite`.  This addresses issue https://github.com/watson-developer-cloud/java-sdk/issues/574.

- It adds the `addCorpus()` method to replace the `addTextToCustomizationCorpus()` method to match the service and the Node SDK.  It deprecates the old version of the method.  It also moves the `allowOverwrite` parameter to the end of the method signature, since it's the sole optional parameter.  This addresses issue https://github.com/watson-developer-cloud/java-sdk/issues/573.

The Speech to Text tests and examples have been modified to use the new version of the method.  I have tested the old version of the method independently of the SDK, and it still works fine.

### DIFF
--- a/examples/java/com/ibm/watson/developer_cloud/speech_to_text/v1/CustomizationExample.java
+++ b/examples/java/com/ibm/watson/developer_cloud/speech_to_text/v1/CustomizationExample.java
@@ -49,7 +49,7 @@ public class CustomizationExample {
 
     try {
       // Add a corpus file to the model:
-      service.addTextToCustomizationCorpus(id, "corpus-1", false, new File(CORPUS_FILE)).execute();
+        service.addCorpus(id, "corpus-1", new File(CORPUS_FILE), false).execute();
 
       // Get corpus status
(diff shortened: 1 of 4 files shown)


germanattanasio:
@jeffpk62 I think you have some JavaDoc problems with java8.

germanattanasio:
Look at Travis

jeffpk62:
Thanks, German!  I didn't check it this time, of course.  I'll get it now.

Is there any way to make these issues show up with `gradle test` or `gradle check`?  That way I could catch them before I commit.

jeffpk62:
Fixed the @link of the deprecated reference to refer to the full signature for the `addCorpus()` metod.

jeffpk62:
Made small parameter changes to clarify the transience of the custom model that's created by the STT integration test.  This should complete issue https://github.com/watson-developer-cloud/java-sdk/issues/503.

Issue #534, opened 2017-01-22
[text-to-speech] getVoice() method not returning custom model information


jeffpk62:
The overloaded TTS `getVoice()` method offers two versions:

- `ServiceCall<Voice> getVoice(String voiceName)`
- `ServiceCall<Voice> getVoice(String voiceName, String customizationId`)

However, they're both returning the same information.

**Without a custom model:**

```
    System.out.println("\nGet voice without custom model...");
    Voice voice = service.getVoice("en-US_AllisonVoice").execute();
    System.out.println(voice);
```

```
{
  "description": "Allison: American English female voice.",
  "gender": "female",
  "language": "en-US",
  "name": "en-US_AllisonVoice",
  "supported_features": {
    "custom_pronunciation": true,
    "voice_transformation": true
  },
  "url": "https://stream.watsonplatform.net/text-to-speech/api/v1/voices/en-US_AllisonVoice"
}
```

**With a custom model:**

```
    System.out.println("\nGet voice with custom model...");
    voice = service.getVoice("en-US_AllisonVoice", custom_id1).execute();
    System.out.println(voice);
```

```
{
  "description": "Allison: American English female voice.",
  "gender": "female",
  "language": "en-US",
  "name": "en-US_AllisonVoice",
  "supported_features": {
    "custom_pronunciation": true,
    "voice_transformation": true
  },
  "url": "https://stream.watsonplatform.net/text-to-speech/api/v1/voices/en-US_AllisonVoice"
}
```

jeffpk62:
This issue is a duplicate of https://github.com/watson-developer-cloud/java-sdk/issues/500.

PR #577, opened 2017-02-16
Improve how Word.Sort enum is handled and general enum comments

jeffpk62:
Changed the `Word.Sort` enum definition to improve how we access the strings associated with the different constants.  Got the idea from how Tradeoff Analytics defines `Column` types.  (And I really disliked the switch statement in the `getWords()` method.)

Also improved the comments for the enums `Word.Type`, `Word.Sort`, and `Customization.WordTypeToAdd`, and tweaked the definition of the IT custom model.


jeffpk62:
The PR also now enables the `customization_id` parameter of the Text to Speech `getVoice()` method.  This addresses issue https://github.com/watson-developer-cloud/java-sdk/issues/534.

Issue #540, opened 2017-01-26
[speech-to-text] Customization does not work when using recognizeUsingWebSocket() method


ota-sandr:
#### Steps to reproduce
```Java
// 1) create RecognitionOptions object with customization ID
RecognizeOptions sttOptions = new RecognizeOptions.Builder()
    .customizationId(MY_CUSTOMIZATION_ID).build();
// 2) use websocket interface with these options
sttService.recognizeUsingWebSocket(audioSource, sttOptions, resultCb);
```

#### Expected behavior

Recognize results should contain the sentences from customization corpus (including custom words) more likely.

#### Actual behavior

When I printed the response message from speech-to-text, it contains string:
```
"warnings": [
   "Unknown arguments: customization_id."
]
```
Recognized results also do not contain custom words.

**JDK version:** 1.8.0_112
**java-skd version:** 3.5.2

#### Suggested solution

Modify WebSocketManager.java :
```Java
public class WebSocketManager {
    // ...
    private class SpeechToTextWebSocketListener implements WebSocketListener {
        private static final String CUSTOMIZATION = "customization_id";
        // ...
        private String buildStartMessage(RecognizeOptions options) {
            JsonObject startMessage = new JsonParser().parse(new Gson().toJson(options)).getAsJsonObject();
            startMessage.remove(MODEL);
            startMessage.remove(CUSTOMIZATION);
            startMessage.addProperty(ACTION, START);
            return startMessage.toString();
        }
        // ...
    }
    // ...
    private WebSocketCall createConnection(RecognizeOptions options) {
        String speechModel = options.model() == null ? "" : "?model=" + options.model();
        if (options.customizationId() != null) {
            speechModel += "&customization_id=" + options.customizationId();
        }
        Builder builder = new Request.Builder().url(url + speechModel);
        // ...
    }
    // ...
}
```

jeffpk62:
Thanks a lot for identifying the problem _and_ for providing the solution!  Made things pretty easy.  Should be fixed once the PR is merged.

PR #581, opened 2017-02-19
Enable customization for Speech to Test WebSocket recognition requests

jeffpk62:
Move the `customization_id` from the `start` message to a query parameter of the connection request for a WebSocket `recognize` call.  This addresses issue https://github.com/watson-developer-cloud/java-sdk/issues/540.


jeffpk62:
@germanattanasio Just let me know what that is and where to do it, and I'll take care of it.

PR #591, opened 2017-02-21
add code coverage to gradle

Mikemosca:
1) Add code coverage to gradle
2) Add the ability to aggregate all the test results in each sub
project to 1 report


Mikemosca:
@germanattanasio - So i did not put the travis changes in because I did not know how to test travis outside of the pull request.  I did not want not break anything.  But the travis changes are minimal.  I think we just need to add 2 entries as follows

script:
  - ./gradlew codeCoverageReport
after_success:
  - bash <(curl -s https://codecov.io/bash)

germanattanasio:
I'm going to wait before I merge this but this looks good! thanks @Mikemosca

PR #593, opened 2017-02-22
Automatic releases

germanattanasio:
Upload artifacts to sonatype for each commit.


.travis.yml
```diff
@@ -36,7 +39,7 @@ script:
 after_success:
   - bash <(curl -s https://codecov.io/bash)
   - .utility/push-javadoc-to-gh-pages.sh
-  - .utility/deploy_snapshot.sh
+  - .utility/deploy.sh
```
germanattanasio: Can we use `deploy:` instead of `after_success` ?
See: https://docs.travis-ci.com/user/deployment/script/

```yml
deploy:
  provider: script
  script: .utility/deploy.sh
  on:
    branch: master
    java: 'oraclejdk7'
```

.utility/deploy.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+#if [ "$TRAVIS_BRANCH" = 'master' ] && [ "$TRAVIS_PULL_REQUEST" == 'false' ]; then
+    openssl aes-256-cbc -K $encrypted_bdb9d73a7940_key -iv $encrypted_bdb9d73a7940_iv -in cd/secring.gpg.enc -out cd/secring.gpg -d
+    ls -l cd
```
germanattanasio: ????

.utility/travis.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
germanattanasio: I don't think we need this file anymore

.travis.yml
blakesteve: Sure thing

.utility/deploy.sh
blakesteve: Good catch, that was still hanging around from debugging. removed

.utility/travis.sh
blakesteve: removed

Issue #536, opened 2017-01-22
[text-to-speech] saveCustomVoiceModel() allows change to language of existing model


jeffpk62:
The TTS customization `saveCustomVoiceModel()` method allows the language of an existing model to be changed.  This functionality is not supported by the HTTP REST method that updates a custom model.  The Java method needs to prevent changing the language for an existing model.

PR #595, opened 2017-02-22
Prevent change of language for existing Text to Speech custom voice model

jeffpk62:
The Text to Speech service does not provide an interface to change the language of an existing custom voice model, which can cause problems with existing word translations that include language-specific pronunciations.  This PR checks whether the `saveCustomVoiceModel()` method is attempting to change the language of an existing model and throws an `IllegalArgumentException` with an appropriate message if so.  Because this changes the way the method works, it could cause failures for users who are relying on the unsupported functionality.  But users really should not change the language of an existing model, so this prevents potential problems for such users.  The PR addresses issue https://github.com/watson-developer-cloud/java-sdk/issues/536.


text-to-speech/src/main/java/com/ibm/watson/developer_cloud/text_to_speech/v1/TextToSpeech.java
```diff
@@ -275,6 +274,26 @@ public TextToSpeech(String username, String password) {
...
+
+          if (existing != null) {
+            Validator.isTrue(model.getLanguage().equals(existing.getLanguage()),
+                             "cannot change language of existing model");
+          }
+
+          return null;
+        }
+      }).execute();
+    }
```
germanattanasio: What's the error that the user will get if they try to change the language? I feel like you are catching the error in the application but it should be the service the one returning this error.

jeffpk62:
@germanattanasio the service doesn't provide a means of changing the language of an existing model, so there's no equivalent error that we can return.  With the service API, once you create a model, its language is immutable.  The problem with the Java SDK is the combination of the create and update functions into a single method.  The best that we can do is prevent the attempt to change the language of an existing model to avoid problems.  I allow the user to pass the same language for a model, since that's not a problem and doesn't require that they remove the language from a local copy of the model before using it as the argument for an update request; it's only a language change that fails.

The best solution is to create separate methods for the create and update operations to mimic the service and prevent the possibility altogether.  But I first wanted to be sure that the SDK's existing method doesn't allow the unsupported operation.

text-to-speech/src/main/java/com/ibm/watson/developer_cloud/text_to_speech/v1/TextToSpeech.java
jeffpk62: @germanattanasio the service doesn't provide a means of changing the language of an existing model, so there's no equivalent error that we can return.  With the service API, once you create a model, its language is immutable.  The problem with the Java SDK is the combination of the create and update functions into a single method.  The best that we can do is prevent the attempt to change the language of an existing model to avoid problems.  I allow the user to pass the same language for a model, since that's not a problem and doesn't require that they remove the language from a local copy of the model before using it as the argument for an update request; it's only a language change that fails.

The best solution is to create separate methods for the create and update operations to mimic the service and prevent the possibility altogether.  But I first wanted to be sure that the SDK's existing method doesn't allow the unsupported operation.

PR #596, opened 2017-02-22
Send audio after sending start message

germanattanasio:
During a call with @daniel-bolanos, he mentioned that we can send the audio right after creating the connection without having to wait for `{ "state": "listening" }`. This Pull Request does that.


daniel-bolanos:
yeah, right after sending the start message... no need to wait for the listening message.

Issue #598, opened 2017-02-23
Add support for addWords() and addWord() methods


jeffpk62:
Add new `addWords()` and `addWord()` methods to the Text to Speech customization API.  This is an enhancement, but it parallels other changes and makes the interface match the service completely.  It's also fairly simple, so we may as well complete the steps to normalize the interfaces.

Issue #531, opened 2017-01-22
[text-to-speech] saveCustomVoiceModel() should be two methods


jeffpk62:
The TTS customization interface for the Java SDK conflates the create and update methods offered by the service for a custom voice model into a single method, `saveCustomVoiceModel()`.  This is basically inconsistent with the service, but a more important inconsistency concerns the inability to add words when updating an existing model.  The update function provided by the service's `POST /v1/customizations/{customization_id}` method lets the user add words to a model that is being updated.  This functionality isn't available with the single combined SDK method.

To be parallel, we should have `createCustomVoiceModel()` to add a new model with no words, and `updateCustomVoiceModel()` that allows both changes to the custom model and the addition of new words.  In this case, the consistency across APIs (HTTP REST, Node.js) is likely worth the additional method.

jeffpk62:
In addition to being more parallel with the service and the other SDKS and to enabling the addition of words on an update, creating two methods may be the best way to preclude allowing users to change the language of an existing model, which should not be possible.

jeffpk62:
Adding support for new words with an update request requires changes related to https://github.com/watson-developer-cloud/java-sdk/issues/533.  For now, we can add the new methods and then address the word support all at once across the API.

Issue #533, opened 2017-01-22
[text-to-speech] Customization getCustomVoiceModel() not returning custom words


jeffpk62:
The TTS `getCustomVoiceModel(id)` method should return the custom words for the single specified model in addition to the information about the model.  The Java method returns the following output:

```
{
  "customization_id": "d4670a06-4dd6-43ee-a943-61c392d776c3",
  "name": "First cURL Test Update",
  "description": "First customization test via cURL update",
  "language": "en-US",
  "owner": "35df7157-fad0-498c-912c-6e429f5c1e11",
  "created": "2017-01-22T16:35:31.128",
  "last_modified": "2017-01-22T16:35:32.471"
}
```
The equivalent HTTP REST and Node.js methods return the following output:

```
{
  "owner": "35df7157-fad0-498c-912c-6e429f5c1e11",
  "customization_id": "0652e868-d141-4df0-a577-ca5f9660046d",
  "created": "2017-01-22T16:44:43.952Z",
  "name": "First cURL Test Update",
  "words": [
    {
      "translation": "ackles",
      "word": "ACLs"
    },
    {
      "translation": "<phoneme alphabet=\"ibm\" ph=\"tr1Ipxl.1i\"></phoneme>",
      "word": "EEE"
    },
    {
      "translation": "<phoneme alphabet=\"ibm\" ph=\"1Y.tr1Ipxl.1i\"></phoneme>",
      "word": "IEEE"
    },
    {
      "translation": "N C double A",
      "word": "NCAA"
    },
    {
      "translation": "I phone",
      "word": "iPhone"
    }
  ],
  "description": "First customization test via cURL update",
  "language": "en-US",
  "last_modified": "2017-01-22T16:45:04.970Z"
}
```

PR #599, opened 2017-02-24
Update Text to Speech customization interface

jeffpk62:
Update the Text to Speech customization interface to become more parallel with the service and with the Node SDK while maintaining the object-based approach to calling methods:

- Add `createCustomVoiceModel()` and `updateCustomVoiceModel()` methods, and deprecate `saveCustomVoiceModel()` method.  The new methods also prevent updating an existing model's language, and they make it possible to provide the capability to add words on an update.  This addresses issue https://github.com/watson-developer-cloud/java-sdk/issues/531.

- Add `addWords()` and `addWord()` method, and deprecate `saveWords()` method.  This is purely to make the interface parallel with the service.  This addresses issue https://github.com/watson-developer-cloud/java-sdk/issues/598.

- Add `getCustomVoiceModel()` method that accepts a model object instead of a customization ID.  The version that accepts an ID is preserved.  The new object-based signature brings the method in line with the other methods of the interface, which accept `CustomVoiceModel` objects to specify a model.

Miscellaneous improvements were also made to comments, JavaDoc, and parameter validation, and all tests and examples are updated accordingly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,10 @@ Change Log
  * Fix: Speech to Text: Enable customization_id for recognizeUsingWebSocket() method
  * Fix: Speech to Text: Fix allowOverwrite parameter of addTextToCustomizationCorpus() method
  * Fix: Speech to Text: Fix error field for WordData object
+ * New: Text to Speech: Add createCustomVoiceModel() and updateCustomVoiceModel() methods, deprecate saveCustomVoiceModel() method
+ * New: Text to Speech: Add addWords() and addWord() methods, deprecate saveWords() method
  * New: Text to Speech: Add getWord() method
+ * New: Text to Speech: Add getCustomVoiceModel() method with CustomVoiceModel argument
  * New: Text to Speech: Add deleteWord() method with String argument
  * New: Text to Speech: Add part_of_speech to CustomTranslation object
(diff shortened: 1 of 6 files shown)


jeffpk62:
Not sure why test coverage percentage has dropped, since I actually added tests.  It could be that I removed tests for deprecated methods.  I could easily put those back if you prefer; they all still pass.

germanattanasio:
Don't worry about test coverage now. The process is not stable enough yet.
I will take a look at this early next week jeff.

jeffpk62:
@germanattanasio, I'll send you quick email summary of what to look at.  Most of this is straightforward and non-controversial, even trivial.  It's really the signatures of the create and update methods that you need to consider.  I'll briefly explain the motivation; they're easily reconsidered if you want changes.  Thanks.

jeffpk62:
Add support for custom words to the `CustomVoiceModel` object by adding a list of `CustomTranslation` objects.  Also, rework the `updateCustomVoiceModel()` method to accept a single parameter, a model object, that can include updates to name and description and can also include one or more new custom word translations.  This PR addresses issue https://github.com/watson-developer-cloud/java-sdk/issues/533 and updates the previous commit for issue https://github.com/watson-developer-cloud/java-sdk/issues/531.  (@germanattanasio, this PR is now probably much more in line with your expectations.)

PR #600, opened 2017-02-24
Sonatype Deployment

blakesteve:
Resolves the issue where snapshots were not being deployed to Sonatype


germanattanasio:
2 comments:
 * Can we move the `cd` and the python script to the `.utility` folder?

core/build.gradle
```diff
@@ -15,7 +15,7 @@ apply plugin: 'checkstyle'
 apply plugin: 'eclipse'
 
 version = '3.5.4-SNAPSHOT'
```
germanattanasio: This version was not updated to be 3.5.5.

germanattanasio:
@schen22 can we leave the version as 3.5.4 for now?

schen22:
@germanattanasio - changed back to 3.5.4

PR #618, opened 2017-03-16
Add all supported languages to enum

dkBrazz:
Translation without modelId specified is possible only with Language enum values. But currently there only several languages declared in the enum.
I added to enum all languages from https://gateway.watsonplatform.net/language-translator/api/v2/identifiable_languages response to be able to do all possible translations.


germanattanasio:
Thanks @dkBrazz

PR #617, opened 2017-03-16
Adds Natural Language Understanding and fix a few bugs in the process

germanattanasio:
## Summary 

* Adds support for NLU thanks to @mkistler 
* Adds `android-utils` to `build.gradle`
* Adds `build.gradle` to the examples
* Only push javadoc for the src folder fixes #547 

I will merge and do a release after that I fix #610

PR #619, opened 2017-03-17
NLU generated SDK code

germanattanasio:


germanattanasio:
I copied most of the changes from this PR into #617

PR #622, opened 2017-03-21
[Documentation] Exclude tests from javadoc generation

### LINKED ISSUE #547: Generate javadoc only for src files not test.
As you can see here: http://watson-developer-cloud.github.io/java-sdk/docs/similarity-search/

We are generating JavaDoc to the unit and integration tests. We should exclude those classes.

blakesteve:
### Summary

This PR resolves the defect where javadocs were incorrectly being generated for test classes.

Resolves #547


germanattanasio:
This is an elegant solution @blakesteve, good job!

PR #632, opened 2017-04-02
[Conversation] workspaces and intents CRUD operations

boazdavid:
:new:
add support for new conversation api.
create/ update/ delete/ get/ list of: workspace/intent
includes javadoc, and unit tests.
code style was checked.

these functions of the api were not implemented:
example/ counterExamples


conversation/src/main/java/com/ibm/watson/developer_cloud/conversation/v1/model/workspace/nodes/CreateDialogNode.java
```diff
@@ -0,0 +1,169 @@
...
+   */
+  public Object getMetadata() {
+    return metadata;
+  }
+
+  /**
+   * @param metadata
+   *          The metadata for the dialog node.
+   */
+  public void setMetadata(Object metadata) {
```
germanattanasio: Object (?) I think it should be `Map<String, Object>`

boazdavid: According to Conversation API it is an object. https://www.ibm.com/watson/developercloud/conversation/api/v1/?curl#create_workspace
```
metadata
object
Any metadata that is required by the workspace.
```
I did not find an example of that object.
Do you think we should represent all JSON objects as `Map<String, Object>`?

germanattanasio:
@boazdavid, I'm looking at merging #646 instead of this. The code in that branch was autogenerated and it's aligned to the direction we are targeting with the SDK.
I will talk to @mkistler to make sure we have the tests and Javadoc you wrote here.
Sorry for the late notice.

germanattanasio:
We are going to merge the code from @mkistler later today. I'm going to close this for now.

PR #637, opened 2017-04-04
[discovery] configuration ID fix

### LINKED ISSUE #623: [discovery] Java SDK ignores the specified configurationId when creating a new collection
Steps to reproduce:

```
// Make sure that the configuration exists
    GetConfigurationRequest getConfigurationRequest = new GetConfigurationRequest.Builder(envId, configId).build();
    GetConfigurationResponse getConfigurationResponse = discovery.getConfiguration(getConfigurationRequest).execute();

    System.out.println("Existing configuration: " +
        "Configuration ID: " + getConfigurationResponse.getConfigurationId() +
        "Name: " + getConfigurationResponse.getName()
    );

    CreateCollectionRequest createCollectionRequest = new CreateCollectionRequest.Builder(envId, configId, collectionName).build();
    Discovery discovery = getDiscoveryClient();
    CreateCollectionResponse response = discovery.createCollection(createCollectionRequest).execute();
    System.out.println("Name: " + response.getName() + ", " + response.getCollectionId()  +
        ", config: " + response.getConfigurationId() + 
        ", status: " + response.getStatus()
    );    

```

Creating a new collection using a non-default config ID ends up with a collection that uses the default config instead, ignoring the specified configuration ID.

SDK versions tested with: 5.5.3, 5.7.0

blakesteve:
This PR converts `configurationId` to `configuration_id` and should resolve #623 

Question: should this raise concern about the camel-case naming convention in our pojos, regarding what the watson services expect?


mkistler:
I don't think this is the right fix.  The Java SDK uses camel case for all property names.  When the JSON property name differs, this should be handled via the `@serializedName` annotation.

blakesteve:
good point Mike, will make these changes today.

maniax89:
good catch on this one - the line that needs the `@SerializedName(ConfigurationManager.ID)` to it is here https://github.com/watson-developer-cloud/java-sdk/blob/develop/discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/collection/CreateCollectionRequest.java#L25

blakesteve:
Apologies for the delay on this, I was out sick for the past several days. Just committed with the correct annotations and added a test for it.

Issue #635, opened 2017-04-03
[text-to-speech]  Spaces in text are encoded as +


arthurfabre:
- [ ] Steps to reproduce:
`TSS.synthesize("Hello Bob", Voice.EN_LISA`);

- [ ] Expected behavior
Audio for "Hello Bob"

- [ ] Actual behavior
Audio for "Hello+Bob"

- [ ] JDK version: OpenJDK 1.8.0_121

- [ ] java-sdk version: 3.7.1

With commit 7d9bbd7dc18b80483b73cf25d398d0ed07b9cbbc, to resolve #602, the text is now url-encoded before being passed off to okhttp.

Unfortunately, `RequestUtils.encode` calls `URLEncoder.encode()`, which performs form-encoding instead of %-encoding. okhttp then does proper %-encoding, which results in requests for synthesizing "Hi Bob" becoming:

    https://stream.watsonplatform.net/text-to-speech/api/v1/synthesize?text=Hi%252C2BBob&voice=en-US_LisaVoice&accept=audio/l16;%20rate%3D48000

Issue #602 seems to be caused by the `é` character being encoded as UTF-8 by okhttp (`0xC3 0xA9`) but decoded as ASCII by the backend, hence the `BadRequestException: 'ascii' codec can't decode byte 0xc3` error.

arthurfabre:
`URLEncoder.encode("é")` returns `%C3%A9` as expected.

okhttp then encodes this again, resulting in `%2525C3%2525A9`. I'm not sure why the `%` character is encoded twice, I would have expected it to become `%25C3%25A9`.

Regardless, this is not the correct solution, as the backend service won't be able to decode the `é` without decoding the URL twice.

Weirdly enough, GET or POST requests, with the text encoded in the URL, using UTF-8 in curl work just fine, with an audible accent:
    
    curl -X GET -u "user:pass" --output test-get.wav "https://stream.watsonplatform.net/text-to-speech/api/v1/synthesize?voice=en-US_AllisonVoice&text=H%C3%A9ello" -w "%{http_code}"

Using a POST request with the text encoded as JSON does not result in an error, however the returned audio does not have an audible accent:

    curl -X POST -u "user:pass" --header "Content-Type: application/json" --data '{"text":"Hello"}' --output test-post.wav "https://stream.watsonplatform.net/text-to-speech/api/v1/synthesize?voice=en-US_AllisonVoice" -w "%{http_code}

arthurfabre:
[...]

Issue #602, opened 2017-03-02
[Text-to-Speech] failed with combine of special characters, e.g Spanish + ";" 


josepmao:
- [ ] Steps to reproduce
_**With following code:**_
```
String text = "La guerra es de vital importancia para el Estado; es el dominio de la vida o de la muerte, el camino hacia la supervivencia o la pérdida del Imperio: es forzoso manejarla bien.";
InputStream stream =service.synthesize(**text**, Voice.ES_ENRIQUE, AudioFormat.WAV).execute();
```

Get error:
```
SEVERE: GET https://stream.watsonplatform.net/text-to-speech/api/v1/synthesize?text=La%20guerra%20es%20de%20vital%20importancia%20para%20el%20Estado;%20es%20el%20dominio%20de%20la%20vida%20o%20de%20la%20muerte,%20el%20camino%20hacia%20la%20supervivencia%20o%20la%20p%C3%A9rdida%20del%20Imperio:%20es%20forzoso%20manejarla%20bien.&voice=es-ES_EnriqueVoice&accept=audio/wav, status: 400, error: 'ascii' codec can't decode byte 0xc3 in position 82: ordinal not in range(128)
com.ibm.watson.developer_cloud.service.exception.BadRequestException: 'ascii' codec can't decode byte 0xc3 in position 82: ordinal not in range(128)
	at com.ibm.watson.developer_cloud.service.WatsonService.processServiceCall(WatsonService.java:416)
	at com.ibm.watson.developer_cloud.service.WatsonService$1.execute(WatsonService.java:180)
	at Jmaotest.testTTS(Jmaotest.java:67)
	at Jmaotest.main(Jmaotest.java:26)
```
**_Change second line code to:_**

```
String output = URLEncoder.encode(text, "UTF-8").replace("+", "%20");
InputStream stream =service.synthesize(output, Voice.ES_ENRIQUE, AudioFormat.WAV).execute();
```

It works as expected.

Because in Java-SDK, we use HTTP GET to send request, thus need URL-encoded, while it seems not working fine while in Spanish.

Suggest either improve URL-encoder part, or change to use POST method.

jeffpk62:
@germanattanasio moving to a POST seems like a good first attempt at a fix.  I can look at this tomorrow.

germanattanasio:
great

jeffpk62:
[...]

PR #638, opened 2017-04-04
[text to speech] Special Character failure resolution

blakesteve:
Related to #635

Reverts the encoding so TTS won't read "+" when returning a space.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,2 +1,2 @@
-version=3.7.1
+version=3.7.2-SNAPSHOT
 group = com.ibm.watson.developer_cloud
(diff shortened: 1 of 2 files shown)


josepmao:
@blakesteve @germanattanasio 
I do not think just remove `String encodedText = RequestUtils.encode(text);` can be a proper change here.  It will cause problem in #602 come back again.  I suppose you add the RequestUtils.encode(text) to avoid that, and even in the workaround from #602, it give extra with `.replace("+", "%20")` after called `URLEncoder.encode()`, although looks it is not a proper fix.

should we consider change to use POST?  If keep using GET, agree with #635 at least need to modifying RequestBuilder to escape semicolons properly

blakesteve:
@josepmao with commit 19948d53f3ad93ac7241ba67fccac8d023bea63e the semicolon gets properly encoded before sent to the server, and testing against the string you provided in #602 it resolves your issue.

josepmao:
@blakesteve thanks for the response.  Get it now :-)

PR #657, opened 2017-04-17
STT: Close websocket after all data processed

arthurfabre:
### Summary

The Speech to Text WebSocketManager never attempts to close the websocket used when streaming audio to the server, as it only attempts to do so in `onOpen()`, if `audioThread` isn't `null`, which I think will never be the case.

According to the API docs, after a request is fully processed, the server sends a state message, indicating that it is ready to process a new request. If the thread responsible for sending data to the server has terminated when we receive such a message, the websocket is now closed.


speech-to-text/src/main/java/com/ibm/watson/developer_cloud/speech_to_text/v1/websocket/WebSocketManager.java
```diff
@@ -144,7 +144,15 @@ public void onMessage(WebSocket webSocket, String message) {
...
       } else if (json.has(STATE)) {
+        // A listen state after everything has been sent over
+        // indicates everything has been processed
+        if (!audioThread.isAlive()) {
+          socket.close(CLOSE_NORMAL, "Transcription completed");
+          return;
+        }
+
         // notify that the service is ready to receive audio
         callback.onListening();
```
germanattanasio: I don't think we should call `onListening();` if we are going to close the socket.

We should do:
```java
if (!audioThread.isAlive()) {
    socket.close(CLOSE_NORMAL, "Transcription completed");
    return;
} else {
    callback.onListening();
}
```

arthurfabre: I don't think we will due to the `return` after `socket.close()`

germanattanasio:
I had to revert this PR because there was a multithreading issue. If Thread-1 sends the audio before the first `onListening()` Then we will close the connection before getting any results.

arthurfabre:
Ah yes, hadn't thought of that.
Would switching to closing the connection upon receiving `final` transcription results work? The STT API supports sending additional requests using the same websocket, but this doesn't seem to be exposed in Java.

germanattanasio:
I actually liked the proposal @arthurfabre if you can rework it so that it doesn't close the WebSocket before getting a transcription

PR #664, opened 2017-04-20
Add support for workspace, intent, example, and counterexample APIs

### LINKED ISSUE #652: [conversation] Add support for workspace, intent, example, and counterexample APIs
The 2017-02-03 version of the Conversation service provides APIs for workspaces, intents, examples, and counterexamples.  These should be added to the Java SDK.

mkistler:
### Summary

Fixes #652

### DIFF
--- a/conversation/src/main/java/com/ibm/watson/developer_cloud/conversation/v1/ConversationService.java
+++ b/conversation/src/main/java/com/ibm/watson/developer_cloud/conversation/v1/ConversationService.java
@@ -13,14 +13,28 @@
 package com.ibm.watson.developer_cloud.conversation.v1;
 
 import com.google.gson.JsonObject;
+import com.ibm.watson.developer_cloud.conversation.v1.model.CounterexampleCollectionResponse;
+import com.ibm.watson.developer_cloud.conversation.v1.model.CreateExample;
+import com.ibm.watson.developer_cloud.conversation.v1.model.CreateWorkspace;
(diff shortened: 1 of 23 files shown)


boazdavid:
all `listXXX()` methods get a `cursor` parameter.
the cursor can be used to get the next page of items.
currently, there is no simple way to extract the cursor from the `pagination.nextUrl()`.
here is the method I used to extract the cursor. I propose to augment the SDK with this functionality:
```
  private String extractCursor(String nextUrl) throws Exception {
    if (nextUrl == null)
      return null;
    final String CURSOR_PARAM = "cursor";
    String query = nextUrl.split("\\?")[1];
    String[] params = query.split("&");
    for (String param : params) {
      String[] pair = param.split("=");
      String key = URLDecoder.decode(pair[0], "UTF-8");
      if (CURSOR_PARAM.equals(key))
        return URLDecoder.decode(pair[1], "UTF-8");
    }
    return null;
  }
```

boazdavid:
`WorkspaceExportResponse.getStatus()` returns String. Consider defining it as an enum

mkistler:
We are working to get support for listXXX() pagination into the next rev of the SDK.  There is some debate on the best way to accomplish this, but I hope we get that resolved soon and get the support into the SDK in the near future.

Agreed that Workspace status should be an enum rather than a string.  We will also look to make this change but that will have some compatibility considerations.

PR #671, opened 2017-04-24
Fix 668

### LINKED ISSUE #668: [Conversation] getCursor()
all listXXX() methods get a cursor parameter.
the cursor can be used to get the next page of items.
currently, there is no simple way to extract the cursor from the response.getPagination.nextUrl().
here is the method I used to extract the cursor. I propose to augment the SDK with this functionality:
```
  private String extractCursor(String nextUrl) throws Exception {
    if (nextUrl == null)
      return null;
    final String CURSOR_PARAM = "cursor";
    String query = nextUrl.split("\\?")[1];
    String[] params = query.split("&");
    for (String param : params) {
      String[] pair = param.split("=");
      String key = URLDecoder.decode(pair[0], "UTF-8");
      if (CURSOR_PARAM.equals(key))
        return URLDecoder.decode(pair[1], "UTF-8");
    }
    return null;
  }
```

### LINKED ISSUE #670: [Conversation] Date deserialization
When I use curl to get workspaces 
https://gateway.watsonplatform.net/conversation/api/v1/workspaces/?version=2017-02-03 
I get the dates in this format:
`"created": "2017-04-23T19:09:46.712Z"`
which is UTC.

But when using the SDK, the date is: `Sun Apr 23 19:09:46 IDT 2017`. in my time zone (Israel, UTC +3). 
Seems that the SDK does not understand the `Z` (UTC) syntax

see also #625

germanattanasio:
* Fixes #668 [Conversation] getCursor()
 * Remove unused variables from Conversation tests
 * Add another date format to try to fix #670

### DIFF
--- a/conversation/src/main/java/com/ibm/watson/developer_cloud/conversation/v1/model/WorkspaceExportResponse.java
+++ b/conversation/src/main/java/com/ibm/watson/developer_cloud/conversation/v1/model/WorkspaceExportResponse.java
@@ -24,6 +24,28 @@ import com.ibm.watson.developer_cloud.service.model.GenericModel;
  */
 public class WorkspaceExportResponse extends GenericModel {
 
+
+  /**
+   * Classifier Status.
+   */
+  public enum Status {
+
+    /** The available. */
+    @SerializedName("Available") AVAILABLE,
+
+    /** The failed. */
+    @SerializedName("Failed") FAILED,
+
+    /** The non existent. */
+    @SerializedName("Non Existent") NON_EXISTENT,
+
+    /** The training. */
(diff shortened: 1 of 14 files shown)


conversation/src/main/java/com/ibm/watson/developer_cloud/conversation/v1/model/WorkspaceExportResponse.java
```diff
@@ -24,6 +24,28 @@
  */
 public class WorkspaceExportResponse extends GenericModel {
 
+
+  /**
+   * Classifier Status.
```
germanattanasio: 😭

PR #683, opened 2017-05-03
First stab at a fully generated SDK for Discovery-V1.

mkistler:
### Summary

This is a first stab at a fully generated SDK for Discovery-V1.

There are many issues to be discussed / reviewed and resolved.  Among these are:

- API names: use listX vs getX for APIs that return a list of {environments,configurations,etc}
- Builders.  The generated code uses parameters rather than builders in a number of cases that differ from the prior hand-written SDK.  The rules used for the generated code are:
    - “GET” operations will collapse their query parameters into an “Options” model if there any optional query parameters.  Currently only the query parameters are collapsed into the options, unlike the handwritten SDK, which also collapses path parameters.
    - Body parameters are “exploded” to a list of parameters corresponding to the body model properties if there are 3 or fewer properties and all are required.  Otherwise, the body parameter is left as is and it’s model implements a builder.
- File parameters are implemented as `InputStream` and for each file parameter we manufacture a companion “MediaType” parameter (optional) that can specify the media type of the file.  The generated SDK does not create a cache of pre-parsed media types — instead it parses the media type on each invocation.
- Handling of dynamic models needs work.  We can’t include any generic (e.g. List<>) types in a dynamic model because the gson parser can’t handle it.  As a result, I omitted the “aggregations” property from QueryResult, which creates a burden for a developer that wants to use nested aggregations.
- I made many more changes to the Swagger
    - Combined `Configuration` and `NewConfiguration`
    - Combined `Document` and `DocumentRef`
    - Combined `Collection` and `CollectionRef`
    - Very significant changes to the models for query responses
    - Changed the `returnParam` parameter from a `string` to `array`of `string`.
- I left in place some query builder code.  None of the query structure is defined in the swagger so this was the best way I could think of to deal with that

I'll be happy for all feedback!


discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/QueryAggregation.java
```diff
@@ -0,0 +1,185 @@
...
+
+import java.util.List;
+
+import com.google.gson.annotations.SerializedName;
+import com.ibm.watson.developer_cloud.service.model.GenericModel;
+
+/**
+ * An aggregation produced by the Discovery service to analyze the input provided.
+ */
+public class QueryAggregation extends GenericModel {
```
maniax89: this class doesn't make sense at all.
we need to discuss this in more depth outside of comments. just for starters, certain types will have different parameters (`term` has a `field`, `count` (String), and `results` whereas something like `timeslice` would have `field`, `interval` (which is a String, not a Long), and `results`). so you can't just ignore the polymorphism here. you won't be able to auto-generate this from the swagger unless it supports custom type deserialization.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/QueryResponse.java
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * A response containing the documents and aggregations for the query.
+ */
+public class QueryResponse extends GenericModel {
+
+  @SerializedName("matching_results")
+  private Long matchingResults;
+  private List<QueryResult> results;
+  private List<QueryAggregation> aggregations;
```
maniax89: this should also have `List<Passages> passages;` but i realize it isn't in the swagger right now

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/NoticeQueryResponse.java
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * NoticeQueryResponse.
+ */
+public class NoticeQueryResponse extends GenericModel {
+
+  @SerializedName("matching_results")
+  private Long matchingResults;
+  private List<NoticeQueryResult> results;
+  private List<QueryAggregation> aggregations;
```
maniax89: assuming that a `NoticeQueryRequest` takes all the same parameters as `DocumentQueryRequest`, this should also have `List<Passage> passages;` - i realize this  is not in the swagger

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/QueryResult.java
```diff
@@ -0,0 +1,95 @@
...
+  public Double getScore() {
+    return (Double) this.get("score");
+  }
+
+  /**
+   * Gets the status.
+   *
+   * @return the status
+   */
+  public String getStatus() {
```
maniax89: what is status in a query result? remove this

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/QueryResult.java
```diff
@@ -0,0 +1,95 @@
...
+  public void setScore(final Double score) {
+    this.put("score", score);
+  }
+
+  /**
+   * Sets the status.
+   *
+   * @param status the new status
+   */
+  public void setStatus(final String status) {
```
maniax89: remove this

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/QueryTemplate.java
```diff
@@ -0,0 +1,119 @@
...
+package com.ibm.watson.developer_cloud.discovery.v1.model;
+
+import java.util.Date;
+
+import com.ibm.watson.developer_cloud.service.model.GenericModel;
+
+/**
+ * QueryTemplate.
+ */
+public class QueryTemplate extends GenericModel {
```
maniax89: query templates are not released, remove this

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/Field.java
```diff
@@ -0,0 +1,97 @@
...
+ */
+package com.ibm.watson.developer_cloud.discovery.v1.model;
+
+import com.google.gson.annotations.SerializedName;
+import com.ibm.watson.developer_cloud.service.model.GenericModel;
+
+/**
+ * Field.
+ */
+public class Field extends GenericModel {
```
maniax89: I had to custom-write a type adapter otherwise Gson blows up when it tries to deserialize this JSON (specifically, type is not a string always, so the default JSON tokenizer throws `IllegalArgumentExcpetion`)

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/collection/field/FieldAdapterFactory.java
```diff
@@ -1,77 +0,0 @@
...
-import com.google.gson.TypeAdapterFactory;
-import com.google.gson.reflect.TypeToken;
-import com.google.gson.stream.JsonReader;
-import com.google.gson.stream.JsonToken;
-import com.google.gson.stream.JsonWriter;
-
-/**
- * Override default enum handling for poorly formed JSON field types schema.
- */
-public class FieldAdapterFactory implements TypeAdapterFactory {
```
maniax89: im pretty sure we will still need this otherwise Gson will blow up on the default JSON parsing

discovery/src/test/java/com/ibm/watson/developer_cloud/discovery/v1/DiscoveryServiceIT.java
```diff
@@ -131,11 +104,10 @@ public static void setupClass() {
     Assume.assumeFalse("config.properties doesn't have valid credentials.",
         (username == null) || username.equals(PLACEHOLDER));
 
-    GetEnvironmentsRequest getRequest = new GetEnvironmentsRequest.Builder().build();
-    GetEnvironmentsResponse getResponse = discovery.getEnvironments(getRequest).execute();
+    GetEnvironmentsResponse getResponse = discovery.getEnvironments(null).execute();
     for (Environment environment : getResponse.getEnvironments()) {
       // look for an existing environment that isn't read only
-      if (!environment.isReadOnly()) {
+      if (!environment.getReadOnly()) {
```
maniax89: so we can't change it so that getters of booleans read as `is<variable>` instead of `get<variable>`?

discovery/src/test/java/com/ibm/watson/developer_cloud/discovery/v1/DiscoveryServiceTest.java
```diff
@@ -483,124 +421,36 @@ public void createDocumentFromInputStreamWithMediaTypeIsSuccessful() throws Inte
...
+    DocumentAccepted response = discoveryService.addDocument(environmentId, collectionId, null, documentStream,
+        HttpMediaType.APPLICATION_JSON, myMetadata.toString(), null).execute();
     RecordedRequest request = server.takeRequest();
 
     assertEquals(DOCS1_PATH, request.getPath());
     assertEquals(POST, request.getMethod());
     assertEquals(createDocResp, response);
   }
 
-  @Test
```
maniax89: why are we removing all these tests?

discovery/src/test/java/com/ibm/watson/developer_cloud/discovery/v1/DiscoveryServiceTest.java
```diff
@@ -106,11 +76,13 @@
...
-      "/v1/environments/mock_envid/collections/mock_collid/query?filter=field:1&query=field:1"
-          + "&count=5&offset=5&return=field&version=2016-12-16";
+      "/v1/environments/mock_envid/collections/mock_collid/query?version=2016-12-16&filter=field%253A1&offset=5"
+          + "&query=field%253A1&count=5&return=field";
   private static final String Q2_PATH =
-      "/v1/environments/mock_envid/collections/mock_collid/query?aggregation=term(field)"
-          + "&count=10&offset=0&version=2016-12-16";
+      "/v1/environments/mock_envid/collections/mock_collid/query?version=2016-12-16&aggregation=term%2528field%2529";
+
+  private static final Long FREE = 0L;
```
maniax89: i think these are enums in the `createEnvironment` `size` parameter

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/NoticeQueryResult.java
```diff
@@ -0,0 +1,243 @@
...
+  }
+
+  /** The unique identifier of the document. */
+  private String id;
+  /** The confidence score of the result's analysis. Scores range from 0 to 1, with a higher score indicating greater
+   *  confidence. */
+  private Double score;
+  /** Status of the processed document which can be 'OK', 'ERROR' depending on if the document type and tags are
+   * accepted by the service. */
+  private String status;
```
maniax89: where did this come from?

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/Discovery.java
```diff
@@ -76,379 +45,803 @@
...
+   *                        metadata file size is 1 MB. Metadata parts larger than 1 MB are rejected. Example:  ``` {
+   *                        "Creator": "Johnny Appleseed",   "Subject": "Apples" } ```
+   * @param configuration   The configuration to use to process the document. If this part is provided, then the
+   *                        provided configuration is used to process the document. If the `configuration_id` is also
+   *                        provided (both are present at the same time), then request is rejected. The maximum
+   *                        supported configuration size is 1 MB. Configuration parts larger than 1 MB are rejected.
+   *                        See the `GET /configurations/{configuration_id}` operation for an example configuration.
+   * @return the {@link DocumentAccepted} with the response
+   */
+  public ServiceCall<DocumentAccepted> addDocument(String environmentId, String collectionId, String configurationId,
```
maniax89: to me this seems to be super cumbersome in java world + against the rule of 3 parameters that you stated would be collapsed into an options object

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/CreateEnvironmentRequest.java
```diff
@@ -0,0 +1,132 @@
...
+ * CreateEnvironmentRequest.
+ */
+public class CreateEnvironmentRequest extends GenericModel {
+
+  /** Name that identifies the environment. */
+  private String name;
+  /** Description of the environment. */
+  private String description;
+  /** Size of the environment. */
+  private Long size;
```
maniax89: why isn't this an enum `[0,1,2,3]`?

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/Discovery.java
```diff
@@ -76,379 +45,803 @@
...
+  }
+
+  /**
+   * Create a collection.
+   *
+   * @param environmentId the ID of your environment
+   * @param body          Input a JSON object that allows you to add a collection.
+   * @return the {@link Collection} with the response
+   */
+  public ServiceCall<Collection> createCollection(String environmentId, CreateCollectionRequest body) {
```
germanattanasio: We decided to do it alphabetically.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/Discovery.java
```diff
@@ -76,379 +45,803 @@
...
+        builder.query("passages", RequestUtils.encode(String.valueOf(queryOptions.passages())));
+      }
+      if (queryOptions.aggregation() != null) {
+        builder.query("aggregation", RequestUtils.encode(queryOptions.aggregation()));
+      }
+      if (queryOptions.sort() != null) {
+        builder.query("sort", RequestUtils.encode(queryOptions.sort()));
+      }
+      if (queryOptions.returnFields() != null) {
+        builder.query("return", queryOptions.returnFields());
```
germanattanasio: Different services use a different encoding. We discussed this in the  #sdk-generation channel.

Text to speech needs `%20` for spaces but Alchemy uses `+`. That's just one example

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/AggregationResult.java
```diff
@@ -0,0 +1,86 @@
...
+  public List<QueryAggregation> getAggregations() {
+    return aggregations;
+  }
+
+  /**
+   * Sets the key.
+   *
+   * @param key the new key
+   */
+  public void setKey(final String key) {
```
germanattanasio: Some responses are being used as requests in some services like Conversation where you want to set properties in the `Context` object

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/DeleteDocumentResponse.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * DeleteDocumentResponse.
+ */
+public class DeleteDocumentResponse extends GenericModel {
+
+  /** The unique identifier of the document. */
+  @SerializedName("document_id")
+  private String documentId;
+  /** Status of the document. A deleted document has the status deleted. */
+  private String status;
```
germanattanasio: We can generate enums but I think we were not going to do it because status could be enhanced and adding a new status is not a breaking change in the API but it's in the SDK.

We have a similar problem with `models` in Speech to Text where there are a set of predefined models like `narrowBand_US` or `broadband_US` but the Speech team keeps adding new models

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/DeleteConfigurationResponse.java
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * DeleteConfigurationResponse.
+ */
+public class DeleteConfigurationResponse extends GenericModel {
+
+  /** The unique identifier for the configuration. */
+  @SerializedName("configuration_id")
+  private String configurationId;
+  /** Status of the configuration. A deleted configuration has the status deleted. */
+  private String status;
```
germanattanasio: We can generate `Enums` but is that the right thing to do? #sdk-generation

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/FullQueryTemplate.java
```diff
@@ -0,0 +1,100 @@
...
+package com.ibm.watson.developer_cloud.discovery.v1.model;
+
+import java.util.List;
+
+import com.ibm.watson.developer_cloud.service.model.GenericModel;
+
+/**
+ * FullQueryTemplate.
+ */
+public class FullQueryTemplate extends GenericModel {
```
germanattanasio: They are in the Swagger...

discovery/src/test/java/com/ibm/watson/developer_cloud/discovery/v1/DiscoveryServiceIT.java
germanattanasio: This is a good point.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/NoticeQueryResult.java
germanattanasio: Swagger. This is all auto generated

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/Discovery.java
germanattanasio: We were discussing when to use the builder pattern and when to use parameters.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/Discovery.java
maniax89: -_-

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/Discovery.java
maniax89: ok fair - shouldn't this be encoded as well then?  `sort` has the exact same format as `return`

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/AggregationResult.java
maniax89: understood. this object (and many others) are not used as requests though. i think we should respect the `readOnly` property in swagger and not generate setters for them

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/DeleteDocumentResponse.java
maniax89: as discussed in slack, I think this should be all or none. I would lean towards specifying the enum especially if it is in the swagger. we should pick and choose which enums to generate (like language) and which ones to not (like status)

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/FullQueryTemplate.java
maniax89: ya that needs fixed

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/NoticeQueryResult.java
maniax89: the reason I ask is I didn't see a `status` in the `NoticeQueryResponse`:

```
NoticeQueryResult {
id (string, optional),
score (number, optional),
notice_id (string, optional): Identifies the notice. Many notices may have the same ID. This field exists so that user applications can programatically identify a notice and take automatic corrective action. ,
created (string, optional),
document_id (string, optional),
severity (string, optional) = ['warning', 'error'],
step (string, optional),
description (string, optional)
}
```

scot-taylor-ibm:
Just confirming @maniax89 's comments that the Query Template stuff needs to be removed from this Swagger. That feature set is not yet released. If you want to go fishing through the git log, you can probably find the commit where the QT stuff was added from `discovery-next.json`, but it's probably easier to manually remove it from `discovery-v1.json` than to undo an archaeological commit.

Regarding `...src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/CreateEnvironmentRequest.java`, OM has an open item for environment sizes > 3. I think they want it to go to 7 or 8. It doesn't need to be a `Long`; a `Short` will do just fine, I believe.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/Discovery.java
mkistler: The alphabetical ordering is imposed by the base [swagger-codegen](https://github.com/swagger-api/swagger-codegen) generator that is not currently configurable and would not be a small effort to fix.  We've chosen to accept this behavior for now because -- the thinking is -- most developers won't actually "read" the SDK files but rather will use code completion and such, which is not impacted by the order of operations in the SDK file.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/Discovery.java
```diff
@@ -76,379 +45,803 @@
...
+    builder.query(VERSION, versionDate);
+    builder.bodyJson(GsonSingleton.getGson().toJsonTree(body).getAsJsonObject());
+
+    return createServiceCall(builder.build(), ResponseConverterUtils.getObject(Environment.class));
+  }
+
+  /**
+   * Query documents.
+   *
+   * See the [Query Language wiki page](https://github.ibm
```
mkistler: Agreed.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/Discovery.java
mkistler: I agree that we should be encoding this array.  Unfortunately, I think the way to do this is to loop through the array encoding each element individually.  That's kinda ugly, but I suppose we do need to do it.

As for your point about doing encoding in the request handler, I agree that we should move in this direction.  If different services need different encodings, we need to write that code somewhere.  Doing it in the request handler will probably make that more efficient and maintainable.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/Discovery.java
```diff
@@ -76,379 +45,803 @@
...
+   * `{{#default parameter_name Default Value}}` - `Default` logic helper. If `parameter_name` is included, it will
+   * be used. Otherwise, the provided default value will be used. For more details, see the [Query Templating design
+   * page](https://github.ibm.com/Watson-Discovery/discovery-wiki/blob/master/design/query-templating.md).
+   *
+   * @param environmentId the ID of your environment
+   * @param collectionId  the ID of your collection
+   * @param body
+   * @return the {@link FullQueryTemplate} with the response
+   */
+  public ServiceCall<FullQueryTemplate> createQueryTemplate(String environmentId, String collectionId, QueryTemplate
```
mkistler: Will do.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/AggregationResult.java
mkistler: I agree that we omit setters for any property marked as `readOnly`.  I'll make this change.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/DeleteDocumentResponse.java
mkistler: The problem with the `['deleted']` enum in the swagger is that it is not complete.  The status will be `deleted` when the delete worked (and has finished), but I don't think that's a guarantee.  So it might be some other value like `processing`, and then that will be a problem for the gson deserialization -- seeing a value there that it does not expect.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/FullQueryTemplate.java
mkistler: I will remove these from the version of the swagger we are using for codegen.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/LanguageEnrichmentOptions.java
```diff
@@ -0,0 +1,254 @@
...
+public class LanguageEnrichmentOptions extends GenericModel {
+
+  /**
+   * If provided, then do not attempt to detect the language of the input document. Instead, assume the language is
+   * the one specified in this field.  You can set this property to work around `unsupported-text-language` errors.
+   * Supported languages include English, German, French, Italian, Portuguese, Russian, Spanish and Swedish.
+   * Supported language codes are the ISO-639-1, ISO-639-2, ISO-639-3, and the plain english name of the language (e
+   * .g. "russian").
+   */
+  public enum Language {
```
mkistler: There is a `LanguageEnrichmentOptions` model in the swagger.  But it seems that it's only use is in a "$allOf" in `EnrichmentOptions`.  So we should probably excluded it from the generated SDK.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/QueryResult.java
mkistler: done.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/QueryResult.java
mkistler: and done.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/Discovery.java
maniax89: that sucks :-(

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/DeleteDocumentResponse.java
maniax89: I actually think this specific status enum is complete, it either is immediately deleted or it throws an error, there is no processing state to my knowledge

discovery/src/test/java/com/ibm/watson/developer_cloud/discovery/v1/DiscoveryServiceIT.java
mkistler: We can change that.  In fact, we had it that way back in March but it got lost when I added support for the Builder pattern in models.  I'll add that back.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/NoticeQueryResult.java
mkistler: It was inherited from `QueryResult` -- but shouldn't have been there either.  It is removed from there and thus will disappear from here.

discovery/src/test/java/com/ibm/watson/developer_cloud/discovery/v1/DiscoveryServiceTest.java
mkistler: Those tests used a `File` parameter rather than `InputStream`.  The variant of the API that accepts `File` is marked "deprecated", so I didn't want to write code to support it only to drop it in the near future.  So instead I just dropped the tests that specify a file.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/Discovery.java
mkistler: Currently we only collapse _query_ parameters for _GET_ method APIs.  Admittedly that was probably short-sighted, but it was focused on fixing the unwieldy `query` method.  I will work on generalizing this to additional parameter types and http methods.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/DeleteDocumentResponse.java
mkistler: If that's true, then it really doesn't matter if this is a string or an enum, because the developer knows that the document is deleted if the response is not an error response.  The developer has no reason to ever look at this field.  In other words -- an enum with one value is just a constant, so this field effectively conveys no information.

That said, the SDK generator can produce the enum if that's really what we want.  I actually had to remove the enum from the swagger to prevent it from doing that.  I'm happy to do whatever we decide is the right thing.

mkistler:
Closing this PR.  To be reopened from new branch.

PR #700, opened 2017-05-15
Improved fully generated SDK for Discovery-V1.

mkistler:
Fully generated Discovery SDK.  Resolves many of the issues raised in initial attempt.

 -Methods that return collections are now named "list<object>".
 -All methods now take an "options" parameter with a builder to set the various fields.
 -String enums are generated as interfaces with string constants -- supporting named values while allowing forward compatibility.

This PR will break backward compatibility in a major way, so we should hold off merging it until we are ready to move to a new major SDK version.  The main goal for this PR is just to get feedback on whether there are additional changes needed.


discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/Discovery.java
```diff
@@ -10,445 +10,721 @@
...
+
+  /**
+   * Queries for notices (errors or warnings) that may have been generated by the system. Currently, notices are only
+   * generated when ingesting documents. See the [Discovery service documentation](https://www.ibm
+   * .com/watson/developercloud/doc/discovery/using.html) for more details on the query language.
+   *
+   * @param queryNoticesOptions the {@link QueryNoticesOptions} containing the options for the call
+   * @return the {@link NoticeQueryResponse} with the response
+   */
+  public ServiceCall<NoticeQueryResponse> queryNotices(QueryNoticesOptions queryNoticesOptions) {
```
maniax89: i wonder how much we benefit from having a separate object for the same exact options as Query - i guess it is good from a point where if they diverge at any point we are prepared for it.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/Collection.java
```diff
@@ -0,0 +1,184 @@
...
+
+/**
+ * A collection for storing documents.
+ */
+public class Collection extends GenericModel {
+
+  /**
+   * The status of the collection.
+   */
+  public interface Status {
```
maniax89: so we decided to use `interface` instead of `enum`?

mkistler: Handling of enums is still open for discussion, but I chose to use interface in this pass because it allows developers to used defined values rather than strings but does not constrain the field to a fixed/never-changing set of values.  This seems like a good compromise that helps developers while allowing for forward compatibility.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/Discovery.java
mkistler: I agree that this seems a little odd -- but it's part of the swagger so I'm assuming it's there for a reason.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/Discovery.java
```diff
@@ -10,445 +10,721 @@
...
+   * | 32,000   | |  **Note:** you cannot set the size property when using the free plan.
+   *
+   * @param createEnvironmentOptions the {@link CreateEnvironmentOptions} containing the options for the call
+   * @return the {@link Environment} with the response
+   */
+  public ServiceCall<Environment> createEnvironment(CreateEnvironmentOptions createEnvironmentOptions) {
+    RequestBuilder builder = RequestBuilder.post("/v1/environments");
+    builder.query(VERSION, versionDate);
+    final JsonObject contentJson = new JsonObject();
+    if (createEnvironmentOptions.size() != null) {
```
germanattanasio: `createEnvironmentOptions` could be null and then you are trying to call `size()`

germanattanasio:
We need to transform the markdown from swagger into HTML. Otherwise, the Javadoc will look pretty ugly. This is actually a cool feature that we could add to `swagger-codegen`.

I saw that when you use `InputStream` for files you specify `filename` as the file name in the header of the request body. Some services( I don't know about Discovery) will use the file extension to know which file is it. We could have a custom method that only is applied when the type is `file`

```java

    /**
     * Sets the image.
     *
     * @param imageStream the image input stream
     * @param imageName the image name
     * @return the builder
     */
    public Builder image(InputStream imageStream, String imageName) {
      Validator.notNull(imageStream, "'imageStream' cannot be null");
      Validator.notNull(imageName, "'imageName' cannot be null");
      this.imageStream = imageStream;
      this.imageName = imageName;
      return this;
    }
```

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/Discovery.java
mkistler: Agreed -- we need to check for createEnvironmentOptions != null above this.  Good catch.

mkistler:
We add a "mediaType" parameter for each file parameter to specify the type of file.  It is currently generated as an independent field of the options file, but if its important we could tie it more closely to the file parameter, e.g. having a builder method that takes both parameters.

germanattanasio:
We could also rename `DiscoveryServiceTest` and `DiscoveryServiceIT` to be `DiscoveryTest` and `DiscoveryIT`

germanattanasio:
Here is a library we could use to transform the markdown into html.

```xml
<dependency>
    <groupId>com.atlassian.commonmark</groupId>
    <artifactId>commonmark</artifactId>
    <version>0.9.0</version>
</dependency>
```
```
Parser parser = Parser.builder().build();
Node document = parser.parse("This is *Sparta*");
HtmlRenderer renderer = HtmlRenderer.builder().build();
renderer.render(document);  // "<p>This is <em>Sparta</em></p>\n"
```

See: https://github.com/atlassian/commonmark-java

mkistler:
Closing this PR.  To be reopened from new branch.

PR #720, opened 2017-06-06
Codegen/discovery

mkistler:
### Summary

Generated SDK with updates to discovery.

### Other Information

Based on discovery-v1.json in the api-explorer repo, codegen/discovery-v1 branch,
commit e7f12a4ff4, modified to remove configurationPart parameter on addDocument,
updateDocument, and testConfigurationInEnvironment.

Generated by watson-swagger-codegen, mdk/java branch, commit 404cc259dbfd4c3b.


germanattanasio:
@mkistler, Andrew was on vacation, he said he will look at this PR this week.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/AggregationResult.java
```diff
@@ -0,0 +1,86 @@
...
+import com.google.gson.annotations.SerializedName;
+import com.ibm.watson.developer_cloud.service.model.GenericModel;
+
+/**
+ * AggregationResult.
+ */
+public class AggregationResult extends GenericModel {
+
+  /** Key that matched the aggregation type. */
+  private String key;
```
maniax89: key could be a `Number` as well when you have a histogram aggregation result

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/QueryAggregation.java
```diff
@@ -0,0 +1,185 @@
...
+
+import java.util.List;
+
+import com.google.gson.annotations.SerializedName;
+import com.ibm.watson.developer_cloud.service.model.GenericModel;
+
+/**
+ * An aggregation produced by the Discovery service to analyze the input provided.
+ */
+public class QueryAggregation extends GenericModel {
```
maniax89: what about `private String path;` for `nested` aggregation? or `private Long count;` for `term` aggregation or `hits` for `top_hits` aggregation?

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/QueryAggregation.java
```diff
@@ -0,0 +1,185 @@
...
+  private List<AggregationResult> results;
+  /** The match the aggregated results queried for. */
+  private String match;
+  /** Number of matching results. */
+  @SerializedName("matching_results")
+  private Long matchingResults;
+  /** Aggregations returned by the Discovery service. */
+  private List<QueryAggregation> aggregations;
+  /** Interval specified by using aggregation type 'timeslice'. */
+  private Long interval;
```
maniax89: when it is a timeslice, `interval` is a `String`, not a `Long` -> when it is a histogram, it is a `Long` - best solution is to make this an `Object` since we aren't going to support polymorphic models

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/QueryNoticesOptions.java
```diff
@@ -0,0 +1,416 @@
...
+      return new QueryNoticesOptions(this);
+    }
+
+    /**
+     * Adds an xreturn to returnFields.
+     *
+     * @param xreturn the new xreturn
+     * @return the QueryNoticesOptions builder
+     */
+    public Builder addXreturn(String xreturn) {
```
maniax89: what is `addXreturn`? can't we make this a better name like `returnFields`?

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/QueryOptions.java
```diff
@@ -0,0 +1,416 @@
...
+      return new QueryOptions(this);
+    }
+
+    /**
+     * Adds an xreturn to returnFields.
+     *
+     * @param xreturn the new xreturn
+     * @return the QueryOptions builder
+     */
+    public Builder addXreturn(String xreturn) {
```
maniax89: can we make this `returnFields` instead of `xreturn`?

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/QueryResult.java
```diff
@@ -0,0 +1,76 @@
...
+  public Double getScore() {
+    return (Double) this.get("score");
+  }
+
+  /**
+   * Gets the metadata.
+   *
+   * @return the metadata
+   */
+  public Map<String, Object> getMetadata() {
```
maniax89: i'm not sure `metadata` is always present on the return object, whereas `id` and `score` are always present - maybe just remove this?

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/Discovery.java
```diff
@@ -10,445 +10,705 @@
...
+    return createServiceCall(builder.build(), ResponseConverterUtils.getObject(Collection.class));
+  }
+
+  /**
+   * Delete a collection.
+   *
+   * @param deleteCollectionOptions the {@link DeleteCollectionOptions} containing the options for the call
+   * @return the service call
+   */
+  public ServiceCall<Void> deleteCollection(DeleteCollectionOptions deleteCollectionOptions) {
```
maniax89: why doesn't this have a response object anymore? it should look something like
```
{
  "collection_id": "my_collection_id",
  "status": "deleted"
}
```

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/Discovery.java
```diff
@@ -10,445 +10,705 @@
...
+   * properties, then they are ignored and overridden by the system, and an error is not returned so that the overridden
+   * fields do not need to be removed when copying a configuration. The configuration can contain unrecognized JSON
+   * fields. Any such fields are ignored and do not generate an error. This makes it easier to use newer configuration
+   * files with older versions of the API and the service. It also makes it possible for the tooling to add additional
+   * metadata and information to the configuration.
+   *
+   * @param createConfigurationOptions the {@link CreateConfigurationOptions} containing the options for the call
+   * @return the {@link Configuration} with the response
+   */
+  public ServiceCall<Configuration> createConfiguration(CreateConfigurationOptions createConfigurationOptions) {
```
maniax89: would have been a lot cleaner to have `CreateConfigurationOptions` to have 2 items in it: environment ID and `Configuration` object that you then do  the `builder.bodyJson(GsonSingleton.getGson().toJsonTree(createConfigurationOptions.configuration())))` on

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/query/Aggregation.java
```diff
@@ -11,7 +11,7 @@
  * specific language governing permissions and limitations under the License.
  */
 
-package com.ibm.watson.developer_cloud.discovery.v1.model.query;
+package com.ibm.watson.developer_cloud.discovery.v1.query;
```
maniax89: are we planning on using these classes or not? seems like the actual service uses `QueryAggregation` instead of this one

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/ListConfigurationsResponse.java
```diff
@@ -10,23 +10,35 @@
...
 
-    public List<Field> getFields() {
-        return fields;
-    }
+  /**
+   * Sets the configurations.
+   *
+   * @param configurations the new configurations
+   */
+  public void setConfigurations(final List<Configuration> configurations) {
```
mkistler: This one is clearly wrong. But I'm not sure we can make the blanket decision to eliminate setters on all response objects.  Need to think about this one.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/AggregationResult.java
mkistler: That's troubling since the swagger says it must be a string:
```
        "key": {
          "type": "string",
          "description": "Key that matched the aggregation type."
        },
```

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/QueryNoticesOptions.java
mkistler: Yes ... we have a way to rename troublesome fields like `return`.  I will use it here.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/Discovery.java
mkistler: We decided to adopt the pattern of returning void for all deleteX methods, since any error on the delete should generate an exception.  @germanattanasio was in favor of this change.  But please state your case if you think this was an unwise decision.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/AggregationResult.java
maniax89: that is unfortunate - what are our options since we cannot do polymorphic models? can we just make it an `Object`?

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/Discovery.java
maniax89: If the API doesn't return anything I figure that is fine, I don't have a strong case for it. I just know that they do have content in the response and it seems unfair to discard it completely - I would ask @jthurne about this one

germanattanasio: If you can come up with a use case where we actually need the response of a `DELETE` method we can talk about having something different than `void`. In all the scenarios I run in my mind I always get non-successful responses as exceptions.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/Discovery.java
germanattanasio: It was cleaner for the sdk developer but it's the same for the Watson user 😉

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/DocumentStatus.java
```diff
@@ -0,0 +1,180 @@
...
+  public List<Notice> getNotices() {
+    return notices;
+  }
+
+  /**
+   * Sets the documentId.
+   *
+   * @param documentId the new documentId
+   */
+  public void setDocumentId(final String documentId) {
```
mkistler: I've tagged all the fields of DocumentStatus as readOnly and fixed the readOnly markings on many other objects/fields.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/Field.java
```diff
@@ -0,0 +1,85 @@
...
+    /** float. */
+    String FLOAT = "float";
+    /** boolean. */
+    String BOOLEAN = "boolean";
+    /** binary. */
+    String BINARY = "binary";
+  }
+
+  private String field;
+  private String type;
```
mkistler: This is a serious problem with the swagger.  Not only is `type` supposed to be a string -- it is also an enum with a supposedly constrained set of values.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/Discovery.java
mkistler: `CreateConfigurationOptions` has a `configuration` setter (this took special effort to create), so the developer can supply just an `environment_id` and `configuration` if that's what he/she wants.  The extra setters on `CreateConfigurationOptions` also make it easy to create a new configuration from an existing one with just one property (e.g. name) different.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/ListConfigurationsResponse.java
mkistler: Created an issue for the SDK generation squad to track this one.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/Field.java
mkistler: Issue opened with the discovery team.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/NormalizationOperation.java
```diff
@@ -0,0 +1,131 @@
...
+    /** merge. */
+    String MERGE = "merge";
+    /** remove. */
+    String REMOVE = "remove";
+    /** remove_nulls. */
+    String REMOVE_NULLS = "remove_nulls";
+  }
+
+  /**
+   * Identifies what type of operation to perform. **copy** - Copies the value of the `source_field` to the
```
mkistler: Agree that we need to do better at this, but we shouldn't hold up this PR for it.  I created an issue in our planning repo: Improve formatting of comments in generated SDKs.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/QueryResult.java
mkistler: It's there because it is present in the swagger.  When it's not provided, the method will just return null, which I think is reasonable behavior.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/query/Aggregation.java
mkistler: I thought these were being used in the tests, but on closer inspection I see that the tests only use the enum classes for query generation.  I'll remove the rest.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/Discovery.java
jthurne: I think returning void here for now is fine. It would be a non-breaking API Change to add a return value in the future.

A possible use-case for returning a status on a delete is to indicate that the delete was unsuccessful, but that you wanted that case to not generate a control-flow-interrupting exception. This is not as unreasonable as it sounds...look at the Java File APIs (both of them). When deleting a file, the delete methods return a boolean indicating if the delete was successful or not.

But Discovery doesn't support this today. If a delete fails, an error HTTP response is returned.

PR #731, opened 2017-06-29
Call configureHttpClient to enable overriding

brad-strauch:
Calling configureHttpClient method to enable overriding


germanattanasio:
@brad-strauch can you sign the CLA?

PR #746, opened 2017-07-09
Codegen/language translator

mkistler:
Update examples and add back Language.java containing string constants for language codes.


germanattanasio:
The failing tests are for conversation

PR #747, opened 2017-07-10
Replace SDK for Personality Insights with generated SDK

mkistler:
Fully generated SDK for Personality Insights V3 with updated tests and supporting files.


personality-insights/src/main/java/com/ibm/watson/developer_cloud/personality_insights/v3/PersonalityInsights.java
```diff
@@ -25,135 +22,115 @@
...
-  }
-
-  /**
-   * Returns a CSV profile.
-   *
-   * @param options the {@link ProfileOptions}
-   * @param includeHeaders if true returns the CSV headers
-   * @return the CSV profile
-   */
-  public ServiceCall<String> getProfileAsCSV(final ProfileOptions options, final boolean includeHeaders) {
```
germanattanasio: The profile could be presented as JSON or CSV. 
We need a method to return either an `InputStream` or a `String` for when the user wants to get a CSV.

mkistler: Hmmm.  This isn't consistent with the policy that we established for Language Translator `identify` operation, which was that if an operation produces multiple formats but one was `application/json`, then we only offer the `application/json` option through the SDKs.  So if we want to do something different here, we need to modify our policy in a way that works for Language Translator `identify` and for Personality Insights `profile`.

Suggestions?

personality-insights/src/main/java/com/ibm/watson/developer_cloud/personality_insights/v3/PersonalityInsights.java
```diff
@@ -25,135 +22,115 @@
...
-    final RequestBuilder requestBuilder = buildProfileRequest(options);
-
-    requestBuilder.header(HttpHeaders.ACCEPT, HttpMediaType.TEXT_CSV);
-    requestBuilder.query(HEADERS, includeHeaders);
-    return createServiceCall(requestBuilder.build(), ResponseConverterUtils.getString());
+    if (profileOptions.contentType().equalsIgnoreCase(ProfileOptions.ContentType.APPLICATION_JSON)) {
+      builder.bodyJson(GsonSingleton.getGson().toJsonTree(profileOptions.content()).getAsJsonObject());
+    } else {
+      builder.bodyContent(profileOptions.body(), profileOptions.contentType());
```
neil-boyette-ibm: To continue on that conversation; doesn't this allow csv responses to be returned?

germanattanasio: The problem is that we are generating the code based on the swagger file @neil-boyette-ibm. If we were writing it by hand it will be easier to just keep this method.

personality-insights/src/main/java/com/ibm/watson/developer_cloud/personality_insights/v3/PersonalityInsights.java
germanattanasio: This is the kind of problem we need to fix either on the API design or the code generation.

PR #759, opened 2017-08-13
Updates to generated SDK for Language Translator

mkistler:
Generated SDK for Language Translator using the latest version of the generator.

The key change in this version of the SDK is that it eliminates the `mediaType` parameters that were previously generated for each file parameter.  This change specifically affects the `forcedGlossary`, `parallelCorpus`, and `monolingualCorpus` parameters of the `createModel` method.  Dropping the mediaType is appropriate because these parameters do not support multiple media types.  This is also consistent with the current hand-written SDK, which does not provide a means to specify a media type for any of these parameters.


language-translator/src/main/java/com/ibm/watson/developer_cloud/language_translator/v2/LanguageTranslator.java
```diff
@@ -115,27 +114,15 @@ public LanguageTranslator(String username, String password) {
     MultipartBody.Builder multipartBuilder = new MultipartBody.Builder();
     multipartBuilder.setType(MultipartBody.FORM);
     if (createModelOptions.forcedGlossary() != null) {
-      MediaType mediaType = null;
-      if (createModelOptions.forcedGlossaryMediaType() != null) {
-        mediaType = MediaType.parse(createModelOptions.forcedGlossaryMediaType());
-      }
-      RequestBody body = InputStreamRequestBody.create(mediaType, createModelOptions.forcedGlossary());
-      multipartBuilder.addFormDataPart("forced_glossary", "filename.tmx", body);  // FIXME: Hand edit to add suffix
+      RequestBody body = InputStreamRequestBody.create(null, createModelOptions.forcedGlossary());
```
mkistler: @germanattanasio I would like your comments specifically on the use of `null` here for the mediaType in `InputStreamRequestBody.create`.  I can confirm that this "works", but if you feel we need to find some way to actually specify a real media type here, we should discuss how best to achieve that.

germanattanasio: Did you manually edit the `filename.tmx` or you are getting that from an annotation?

mkistler: Adding ".tmx" was a hand-edit.  But hopefully this is no longer needed

https://github.ibm.com/mnlp/wdc-tracker/issues/837

PR #772, opened 2017-08-17
Add unit and integration tests

lpatino10:
The new Training Data endpoint is tested. Some models were also changed
to make tests pass, as they were generated wrong. This issue has since
been fixed.


germanattanasio:
@mkistler I'm going to merge this but we need to regenerate the code because @lpatino10 is using `int` instead of `Integer`

PR #779, opened 2017-08-28
Add support for additionalProperties

lpatino10:
### Summary

Models which previously extended `HashMap<String, Object>` should now work with the same methods as those which extend `GenericModel`, while retaining the ability to hold dynamic properties.


mkistler:
This looks good.  But we should break this into changes in the core for DynamicModels and a separate PR to Conversation that uses the new DynamicModels.  The PR to Conversation should be generated, so we'll need to update the generator to produce DynamicModels for models that have `additionalProperties`.

lpatino10:
I can do that. Would it also make more sense to move the DynamicModel class into the core along with ObjectModel and GenericModel?

lpatino10:
I'm closing this to make a new one with the generated files left out.

PR #783, opened 2017-08-31
Add newest generated code and update tests accordingly

lpatino10:
### Summary

Updates were made to the Java SDK generator to handle DynamicModel objects. These changes were used to generate this new code and tests were updated to reflect any model changes.

Some handwritten classes were also edited, like the DynamicModel and ObjectModel, and a new helper class was written to help with serialization of properties of objects extending DynamicModel.

### DIFF
--- a/conversation/src/main/java/com/ibm/watson/developer_cloud/conversation/v1/model/Counterexample.java
+++ b/conversation/src/main/java/com/ibm/watson/developer_cloud/conversation/v1/model/Counterexample.java
@@ -27,7 +27,7 @@ public class Counterexample extends GenericModel {
 
   /**
    * Gets the text.
-   *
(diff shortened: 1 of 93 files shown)


germanattanasio:
@lpatino10 can you do PR from branches in this repo? that way we can run the integration tests.

conversation/src/main/java/com/ibm/watson/developer_cloud/conversation/v1/model/Counterexample.java
```diff
@@ -27,7 +27,7 @@
 
   /**
    * Gets the text.
-   *
+   * <p>
```
germanattanasio: What's this `<p>` doing here?

PR #787, opened 2017-09-05
Update generated Discovery code

lpatino10:
### Summary

New generated code for Discovery, as well as any changes that came from syncing with the develop branch, have been added. Discovery should now work with the DynamicModel implementation used in the Conversation service. A new test has also been added to test the passages parameter of Discovery queries.


lpatino10:
I'm going to make a new PR without all the Conversation changes.

PR #791, opened 2017-09-05
Codegen/discovery

germanattanasio:
- [x] Generated code for discovery
 - [x] Query method

This PR adds models for the various `QueryAggregation` types. Tests have been updated accordingly.


germanattanasio:
I think the last question we had before merging this was if the `query` request and response were being mapped to object or not.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/Collection.java
```diff
@@ -26,43 +26,42 @@
    * The status of the collection.
    */
   public interface Status {
-    /** active. */
+    /**
```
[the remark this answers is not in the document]
lpatino10: Yeah I think it did.

lpatino10:
Here's an example of dealing with aggregations with the new changes:
```java
QueryResponse queryResponse = discovery.query(queryOptions).execute();
Histogram histogram = (Histogram) queryResponse.getAggregations.get(0);
Long interval = histogram.getInterval();
```
This should look the same as it did before we started using generated models. Does this look good @maniax89?

maniax89:
this looks great! that AggregationSerializer is pretty cool.

commentary - seems like a lot of the response classes have setters on them which i feel like they shouldn't, but maybe not the focus of this PR

lpatino10:
Thanks! And yeah we'll have to work together to update the models now that we have better support for this. Getting rid of the setters won't be difficult and I also know there are more aggregations to add in the future, for example.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/AddDocumentOptions.java
```diff
@@ -18,31 +18,33 @@
...
+    /** application/vnd.openxmlformats-officedocument.wordprocessingml.document. */
+    String APPLICATION_VND_OPENXMLFORMATS_OFFICEDOCUMENT_WORDPROCESSINGML_DOCUMENT
+        = "application/vnd.openxmlformats-officedocument.wordprocessingml.document";
+    /** application/pdf. */
+    String APPLICATION_PDF = "application/pdf";
+    /** text/html. */
+    String TEXT_HTML = "text/html";
+    /** application/xhtml+xml. */
+    String APPLICATION_XHTML_XML = "application/xhtml+xml";
+  }
```
mkistler: FYI ... the FileMediaType interface will go away with the changes I am making for file handling.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/QueryNoticesOptions.java
```diff
@@ -147,6 +122,36 @@ public Builder addReturnField(String returnField) {
     }
 
     /**
+     * Adds an sort to sort.
```
mkistler: `sort` is described in the API Ref as
> a comma-separated list of fields in the document to sort on.

So an element of the `sort` array is a `field`.

It we want this to be reflected in the SDK, we just need to add an `x-item-name` annotation on `sort` with the value `"field"`.  Or maybe we would want `sort_field`.  Then the comment above would read
```
Adds an sort_field to sort
```
and the method signature below would be
```
public Builder addSortField(String sortField) {
```

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/QueryNoticesOptions.java
```diff
@@ -147,6 +122,36 @@ public Builder addReturnField(String returnField) {
...
+      if (this.sort == null) {
+        this.sort = new ArrayList<String>();
+      }
+      this.sort.add(sort);
+      return this;
+    }
+
+    /**
+     * Adds an passagesFields to passagesFields.
+     *
```
mkistler: This is another place where we might want to use `x-item-name`, probably just "passagesField".

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/QueryNoticesOptions.java
```diff
@@ -336,6 +399,10 @@ public String filter() {
   /**
    * Gets the query.
    *
+   * A query search returns all documents in your data set with full enrichments and full text, but with the most
+   * relevant documents listed first. Use a query search when you want to find the most relevant search results. You
+   * cannot use `natural_language_query` and `query` at the same time.
+   *
```
mkistler: If we were writing this SDK by hand, we'd validate that `natural_language_query` and `query` are not both specified when constructing the options instance from the builder.  There's no support for this in the generator today but we should consider adding it.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/UpdateConfigurationOptions.java
```diff
@@ -205,18 +195,18 @@ public Builder enrichments(List<Enrichment> enrichments) {
...
-      this.conversions = configuration.getConversions();
-      this.name = configuration.getName();
-      this.description = configuration.getDescription();
-      this.normalizations = configuration.getNormalizations();
-      this.enrichments = configuration.getEnrichments();
+    this.conversions = configuration.getConversions();
+    this.name = configuration.getName();
+    this.description = configuration.getDescription();
+    this.normalizations = configuration.getNormalizations();
+    this.enrichments = configuration.getEnrichments();
```
mkistler: More funky formatting here.  This case looks like some spacing issues in the template.  But odd that it is showing up here and not in the other classes.

mkistler: Oh ... I bet this is one of the "alternate-setters".  Not in the template ... these lines are produced in the generator itself.

PR #812, opened 2017-10-17
Adding customization weight for speech to text

DigitalZebra:
Added customization_weight parameter for speech to text. This is based on documentation here: https://www.ibm.com/watson/developercloud/speech-to-text/api/v1/#recognize_audio_websockets

Let me know if I messed anything up. Thanks!

### DIFF
--- a/speech-to-text/src/main/java/com/ibm/watson/developer_cloud/speech_to_text/v1/SpeechToText.java
+++ b/speech-to-text/src/main/java/com/ibm/watson/developer_cloud/speech_to_text/v1/SpeechToText.java
@@ -235,7 +235,7 @@ public class SpeechToText extends WatsonService {
     }
 
     if (options.customizationWeight() != null) {
-      requestBuilder.query(CUSTOMIZATION_WEIGHT, options.customizationWeight().toString());
+      requestBuilder.query(CUSTOMIZATION_WEIGHT, options.customizationWeight());
     }
   }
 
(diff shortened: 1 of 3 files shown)


speech-to-text/src/main/java/com/ibm/watson/developer_cloud/speech_to_text/v1/SpeechToText.java
```diff
@@ -232,6 +233,10 @@ private void buildRecognizeRequest(RequestBuilder requestBuilder, RecognizeOptio
     if (options.customizationId() != null) {
       requestBuilder.query(CUSTOMIZATION_ID, options.customizationId());
     }
+
+    if (options.customizationWeight() != null) {
+      requestBuilder.query(CUSTOMIZATION_WEIGHT, options.customizationWeight().toString());
```
lpatino10: We're not using `toString()` on the other `Double` parameters, so maybe we should remove that to stay consistent.

speech-to-text/src/main/java/com/ibm/watson/developer_cloud/speech_to_text/v1/SpeechToText.java
```diff
@@ -815,6 +820,10 @@ public WebSocket recognizeUsingWebSocket(final InputStream audio, final Recogniz
       urlBuilder.addQueryParameter(CUSTOMIZATION_ID, options.customizationId());
     }
 
+    if (options.customizationWeight() != null) {
+      urlBuilder.addQueryParameter(CUSTOMIZATION_WEIGHT, options.customizationWeight().toString());
```
lpatino10: Same comment as line 238.

DigitalZebra:
Thanks for the review! I updated both locations where changes were requested. I still had to explicitly convert `Double` to `String` on line 824 because `addQueryParameter` requires a `String` instead of an `Object`. See this line: https://github.com/watson-developer-cloud/java-sdk/pull/812/commits/9862f4ffbc98f8ac9ef8475a1112aea316aabf5b#diff-7cbde1ffca56bf8d42f2a65fdbba42eeR824 

Please let me know if you would like any other changes. Thanks!

PR #821, opened 2017-11-27
Retrieve service URL from JDNI

### LINKED ISSUE #553: Retrieve service url from JDNI
## As a Developer _I want_ to set the URL automatically when running in Bluemix _so that_ I can use the services running in other places rather than public

**Context:**

We currently load the username and password from a JDNI variable: `"watson-developer-cloud/" + serviceType + "/credentials"`. We need to update the method that retrieve that to also set the service URL. from `"watson-developer-cloud/" + serviceType + "/url"`.

### Acceptance Criteria:

 - [ ] Set `url` for services automatically.

lpatino10:
Resolves #553 

This PR adds a default JDNI lookup of the service URL when instantiating a service while slightly refactoring the method for retrieving values from JDNI.


germanattanasio:
@lpatino10 can you add a test case for this?

PR #830, opened 2017-12-06
Make sure Travis updates the GitHub pages on tagged commits

lpatino10:
The GitHub pages weren't being updated on tagged commits. This small change in the Travis YAML file should fix that.


gradle.properties
```diff
@@ -1,2 +1,2 @@
-version=4.1.1
+version=4.0.2-SNAPSHOT
```
germanattanasio: 🤔

PR #832, opened 2017-12-12
Automatic jar upload

### LINKED ISSUE #831: Automatically upload release jars
Upon new releases, the jar is generated but not automatically uploaded to the GitHub releases entry. This functionality should be added.

lpatino10:
Fixes #831 

This PR ensures that upon making releases, the final jar with dependencies is automatically uploaded to the appropriate GitHub release.

It also fixes the missing dependency for one of the new core unit tests, which was previously being skipped in Travis. This should now run properly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,8 +57,8 @@ deploy:
       jdk: oraclejdk7
 
   - provider: releases
-    api_key: $GITHUB_TOKEN_RELEASES
-    file: java-sdk/build/libs/jar-with-dependencies-$TRAVIS_BRANCH.jar
+    api_key: ${GITHUB_TOKEN_RELEASES}
+    file: java-sdk/build/libs/jar-with-dependencies-${TRAVIS_BRANCH}.jar
     skip_cleanup: true
     on:
       repo: watson-developer-cloud/java-sdk
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,5 +56,14 @@ deploy:
       repo: watson-developer-cloud/java-sdk
       jdk: oraclejdk7
 
+  - provider: releases
+    api_key:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -56,5 +56,14 @@ deploy:
       repo: watson-developer-cloud/java-sdk
       jdk: oraclejdk7
 
+  - provider: releases
+    api_key:
+      secure: NhwGz9ppK1YM0M9Pr15b6fOHPpN09QrLvdK4z1PTRxAt+oVpzl3xM/CpGr+iRXT3aV+8q0xLtb4rjxdmgV7aDFY5WEcKkdHAlENpLk/ZVCswnl/MakNbupFbiml0aLcNRe90qvuMILjwSfzYU6BlxZ/W1Xz1eqwtQ1oP/A92MZM=
```
germanattanasio: Use `$GITHUB_TOKEN` so that we can control this value on the Travis settings page. I don't like doing commits to change a secret.

core/build.gradle
```diff
@@ -75,6 +75,7 @@ dependencies {
     compile group: 'com.google.code.gson', name: 'gson', version: '2.8.0'
     compile group: 'org.apache.commons', name: 'commons-lang3', version: '3.5'
     compile group: 'org.glassfish.jersey.bundles.repackaged', name: 'jersey-jsr166e', version: '2.25.1'
+    compile group: 'simple-jndi', name: 'simple-jndi', version: '0.11.4.1'
```
germanattanasio: should this be a test dependency?

germanattanasio:
@lpatino10 I think you need to do
```
file: java-sdk/build/libs/jar-with-dependencies-${TRAVIS_BRANCH}.jar
```

PR #833, opened 2017-12-12
Add handwritten method for getting profile in CSV format

lpatino10:
Before the Personality Insights service was being generated programmatically, there was a method to get the returned profile in CSV format. This was removed in the 4.0.0 release.

This PR adds that functionality back with a new `getProfileAsCSV()` method.


lpatino10:
Yeah if `csvHeaders` doesn't do anything for `profile()`, that might make sense. And since we're currently generating `profile()` and not `getProfileAsCSV()`, we could just remove it from swagger.

For the purpose of this PR I can get rid of those manually.

germanattanasio:
Do we have a wiki page for this kind of handwritten pieces?

lpatino10:
Not that I know of.. I can talk with some people to get that done because I think it would be good information to have out in the open.

PR #834, opened 2017-12-12
Discovery patch

### LINKED ISSUE #826: [Discovery] - Add/update configuration does not work
I am using the latest Java SDK. I tried adding configuration to Discovery using Java API. But it is failing. Although the config is pushed to Discovery, when you try to open the configuration at Discovery UI it does not since the configuration is parsed and uploaded in an incorrect format.

After debugging for a while, I noticed that, when parsing local config Json, it is skipping options part of enrichment. Please look at it and get the issue resolved. We have a client waiting for this feature.

Here is the code I used to add config:
```java
Discovery discovery = new Discovery("Version");
discovery.setEndPoint("https://gateway.watsonplatform.net/discovery/api/");
discovery.setUsernameAndPassword("username:password");
String environmentId = "Env_ID";
String configurationName = "ConfigName";

CreateConfigurationOptions.Builder createBuilder = new CreateConfigurationOptions.Builder(environmentId);
Configuration configuration =  GsonSingleton.getGson().fromJson(
    new FileReader("config.json"), com.ibm.watson.developer_cloud.discovery.v1.model.Configuration.class);
createBuilder.configuration(configuration);
Configuration createResponse = discovery.createConfiguration(createBuilder.build()).execute();
```

lpatino10:
Fixes #826 

The Discovery service has moved to using NLU-based enrichment options. This PR makes those corresponding model changes.

### DIFF
--- a/discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/NluEnrichmentEmotion.java
+++ b/discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/NluEnrichmentEmotion.java
@@ -59,12 +59,12 @@ public class NluEnrichmentEmotion extends GenericModel {
      * @param targets the new targets
      * @return the NluEnrichmentEmotion builder
      */
-    public Builder addTargets(String targets) {
-      Validator.notNull(targets, "targets cannot be null");
+    public Builder addTarget(String target) {
+      Validator.notNull(target, "target cannot be null");
       if (this.targets == null) {
         this.targets = new ArrayList<String>();
       }
-      this.targets.add(targets);
+      this.targets.add(target);
       return this;
     }
 
--- a/discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/NluEnrichmentEmotion.java
+++ b/discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/NluEnrichmentEmotion.java
@@ -54,9 +54,9 @@ public class NluEnrichmentEmotion extends GenericModel {
     }
 
     /**
-     * Adds an targets to targets.
+     * Adds a target to targets.
      *
-     * @param targets the new targets
+     * @param target the new target
      * @return the NluEnrichmentEmotion builder
      */
     public Builder addTarget(String target) {
--- a/discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/NluEnrichmentEmotion.java
+++ b/discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/NluEnrichmentEmotion.java
@@ -12,9 +12,11 @@
  */
 package com.ibm.watson.developer_cloud.discovery.v1.model;
 
+import java.util.ArrayList;
 import java.util.List;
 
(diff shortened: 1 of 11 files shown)


discovery/src/test/java/com/ibm/watson/developer_cloud/discovery/v1/DiscoveryServiceIT.java
```diff
@@ -606,15 +629,30 @@ public void updateConfigurationIsSuccessful() {
...
+    keywords.setSentiment(true);
+    NluEnrichmentSemanticRoles semanticRoles = new NluEnrichmentSemanticRoles();
+    semanticRoles.setEntities(true);
+    NluEnrichmentFeatures features = new NluEnrichmentFeatures();
+    features.setSentiment(sentiment);
+    features.setEmotion(emotion);
+    features.setEntities(entities);
+    features.setKeywords(keywords);
+    features.setSemanticRoles(semanticRoles);
+    EnrichmentOptions options = new EnrichmentOptions();
```
mkistler: I think we probably want builders generated for `EnrichmentOptions`, `NluEnrichmentFeatures`, `NluEnrichmentSemanticRoles`, `NluEnrichmentKeywords`, `NluEnrichmentEntities`, `NluEnrichmentEmotion`, and `NluEnrichmentSentiment`.

lpatino10: Agreed. I'll look at making that change in the generator and updating this.

germanattanasio:
I think this is going to break existing users. 🔥

lpatino10:
Unfortunately I think so too 😞

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/NluEnrichmentEmotion.java
```diff
@@ -0,0 +1,130 @@
...
+      return new NluEnrichmentEmotion(this);
+    }
+
+    /**
+     * Adds an targets to targets.
+     *
+     * @param targets the new targets
+     * @return the NluEnrichmentEmotion builder
+     */
+    public Builder addTargets(String targets) {
```
mkistler: We need `"x-item-name": "target"` on this array so that we get
```
public Builder addTarget(String target) {
```

We could probably make this change by hand here, rather than regenerate the whole SDK -- as long as we make sure to fix the swagger for future iterations.

discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/NluEnrichmentSentiment.java
```diff
@@ -0,0 +1,130 @@
...
+      return new NluEnrichmentSentiment(this);
+    }
+
+    /**
+     * Adds an targets to targets.
+     *
+     * @param targets the new targets
+     * @return the NluEnrichmentSentiment builder
+     */
+    public Builder addTargets(String targets) {
```
mkistler: We need `"x-item-name": "target"` on this array so that we get
```
public Builder addTarget(String target) {
```

We could probably make this change by hand here, rather than regenerate the whole SDK -- as long as we make sure to fix the swagger for future iterations.

PR #842, opened 2018-01-02
Set authorization in header for VR dedicated

### LINKED ISSUE #838: [Visual_Recognition] Issue of authorization in dedicated environment.
Hi.
I found the issue of Authorization in Dedicated environment.

In our Visual Recognition Dedicated, the credentials are username and password. This is different from the public environment. (not API key)
However, we can't authenticate the user with the SDK 'setUsernameAndPassword()' method.
When this method is used, a 401 authentication error occurs.
We also changed the endpoint to our endpoint with the 'setEndPoint()' method.
Attach the code and error log below.

- code (Instance declaration section)

VisualRecognition service = new VisualRecognition( VisualRecognition.VERSION_DATE_2016_05_20);
service.setEndPoint("our endpoint");
service.setUsernameAndPassword("username","password");

- error
status: 401,
error: 
<HTML>
  <BODY>
    <p>Watson Message</p>
    <p>Credentials Required</p>
  </BODY>
</HTML>

lpatino10:
Fixes #838 

~~This PR makes sure that if the user of a VR service changes the endpoint, the authorization is set in the header like in other services.~~

~~Unfortunately, the change is being made to the VR service file, meaning it will have to be tweaked post-codegen. This is just due to the nature of VR using an API key most of the time but a username and password for the dedicated option.~~

After new updates, `WatsonService` has been changed to store the username and password. The VR service will then check for those values to decide which authorization method to use. Presence of a username and password guarantees the use of a dedicated VR service.

The new changes to the VR service class can also be added to the generator code, eliminating the need for post-generation tweaks.

### DIFF
--- a/visual-recognition/src/main/java/com/ibm/watson/developer_cloud/visual_recognition/v3/VisualRecognition.java
+++ b/visual-recognition/src/main/java/com/ibm/watson/developer_cloud/visual_recognition/v3/VisualRecognition.java
@@ -59,6 +59,8 @@ public class VisualRecognition extends WatsonService {
   /** The Constant VERSION_DATE_2016_05_20. */
   public static final String VERSION_DATE_2016_05_20 = "2016-05-20";
 
+  private boolean endPointChanged;
+
   /**
    * Instantiates a new `VisualRecognition`.
    *
@@ -89,6 +91,17 @@ public class VisualRecognition extends WatsonService {
     setApiKey(apiKey);
   }
 
+  /**
+   * Sets the endpoint while making a note of the change to later authenticate properly.
+   *
+   * @param endPoint the new endpoint. Will be ignored if empty or null
+   */
+  @Override
+  public void setEndPoint(String endPoint) {
+    super.setEndPoint(endPoint);
+    endPointChanged = true;
+  }
+
   /*
    * (non-Javadoc)
    */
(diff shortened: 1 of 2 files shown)


visual-recognition/src/main/java/com/ibm/watson/developer_cloud/visual_recognition/v3/VisualRecognition.java
```diff
@@ -97,6 +110,9 @@ protected void setAuthentication(okhttp3.Request.Builder builder) {
     if (getApiKey() == null) {
       throw new IllegalArgumentException("api_key needs to be specified. Use setApiKey()");
     }
+    if (endPointChanged && !getEndPoint().equals(URL)) {
+      super.setAuthentication(builder);
```
mkistler: Would it be wrong to set this authentication header even if the request is targeting a public instance of VR?

If not, then why not just supply it always and keep the code simple.

But if it is a problem, then I worry that the endpoint might be changed because it is for a service instance in a different region.  In that case, we don't want to provide this authorization header.

lpatino10: Yeah we only want to set the authentication header if it's a dedicated instance. Otherwise, the service wants the API key as a query parameter.

You're right though that there might be cases where the endpoint is changed but isn't pointing to a dedicated instance.. Is there any way to check whether or not the user is trying to use one?

mkistler: I think my preference would be to change the WatsonService so that it stores the username/password separately from the APIKey.  If we do this, then we can prefer the username/password whenever it it specified, and fall back to APIKey (for VR) otherwise.

lpatino10: Agreed. I'll look at making that change now.

PR #857, opened 2018-02-03
Add generated code for Speech to Text service

lpatino10:
This PR adds the generated models and service class for Speech to Text.

All unit and integration tests have been updated to reflect the changes, and new ones have been added as necessary as well.

Any manual changes necessary post-generation can be found [here](https://github.ibm.com/MIL/watson-swagger-codegen/wiki/Java-Post-Generation-Notes).


germanattanasio:
We need to make sure the `RecognizeOptions` can be share between the REST and the WebSocket endpoints

PR #863, opened 2018-02-09
Add newest generated code for Natural Language Classifier

lpatino10:
This PR adds new newest generated code for Natural Language Classifier as of 2/8/18 for the purpose of the Java SDK 5.0.0 release.


germanattanasio:
### TODO
- [ ] Document breaking changes
- [ ] Verify that we can create a classifier with an in-memory CSV file.

germanattanasio:
@lpatino10 Can you keep the existing methods at the bottom and adapt them to use the new methods? That way this won't be a breaking change.

Issue #741, opened 2017-07-05
Use Jdiff to publish the API changes along with the javadoc


germanattanasio:
See: http://javadiff.sourceforge.net/

Guava uses this in their release history: https://github.com/google/guava/wiki/ReleaseHistory

germanattanasio:
@chohanbin if the build passes and we publish the API Diff then we can close this issue

lpatino10:
Unfortunately, the changes we made don't quite work yet. Hopefully fixes in #949 will let us close this.

germanattanasio:
I'm looking at http://watson-developer-cloud.github.io/java-sdk/apidiff/ and I don't think it's working 🤔

germanattanasio:
is this working @lpatino10?

lpatino10:
Nope, still not working 😕

PR #869, opened 2018-02-15
Add a Gradle task to generate an API diff report

chohanbin:
### Summary
- This change is an answer issue #741 ("Use Jdiff to publish the API changes along with the javadoc"). But since the ultimate goal is to produce an API diff, I also considered API diff tools other than JDiff as well (considering ease of usage, active support...etc.), which leads to...
- Generating API diff is handled by `japicmp` plugin. (Original japicmp project using maven: https://github.com/siom79/japicmp  Gradle plugin of japicmp, which I am using: https://github.com/melix/japicmp-gradle-plugin)

- Usage:
1) Edit `java-sdk/build.gradle`'s `apidiff` task - to set the proper file paths for two JARs to compare (Allowing the file paths to be passed in as the command-line argument is a future to-do).
2) Invoke `gradle apidiff` from the root `java-sdk/` directory.
3) Check that the reports have been generated in the report file paths specified in `build.gradle`'s `apidiff` task.
Bonus) Check out the API diff report between `java-sdk-4.1.0-jar-with-dependencies.jar `and `java-sdk-4.2.1-jar-with-dependencies.jar`, in the attachment. japicmp can produce HTML, XML, and TXT reports. Unfortunately, GitHub won't let me attach HTML report (with fancy hyperlinks!) in this pull request.

[api_diff.txt](https://github.com/watson-developer-cloud/java-sdk/files/1728494/api_diff.txt)

I'll look for how-to allow the JAR file paths to be passed in as the command-line argument to this Gradle task.

Let me know for any feedback! :)


germanattanasio:
This is a good start. I think you can download the old jar from maven and generate the new one based on the code. The report should be part of the github pages site. We have an script for something like this https://github.com/watson-developer-cloud/java-sdk/blob/develop/.utility/push-javadoc-to-gh-pages.sh

germanattanasio:
Hey @chohanbin, if you can generate the diff in a folder based on the existing commit and the last tag then I can help you publish that using Travis.

PR #867, opened 2018-02-09
Add newest generated code for Visual Recognition

lpatino10:
This PR adds new newest generated code for Visual Recognition as of 2/8/18 for the purpose of the Java SDK 5.0.0 release.

PR #875, opened 2018-02-27
Generated Visual Recognition service using vr-form-params swagger

mkistler:
Generated VR service using a version of the swagger that replaces the "parameters" json-string formData param with separate formData parameters for the values it may contain.

This contains the hand-modifications to create/updateClassifier and updates to the tests.


germanattanasio:
This will be done by @lpatino10 in #867

PR #895, opened 2018-03-14
Fix broken parameter passing when calling recognizeUsingWebSocket

lpatino10:
Currently, when setting the JSON body for a `recognizeUsingWebSocket()` call, the JSON keys are left as camelCase, so some parameters are being unrecognized by the service (ex: `interimResults`).

This PR ensures that the JSON body is properly formatted.


germanattanasio:
Let's release this as a patch release tomorrow morning.

PR #902, opened 2018-03-23
Update READMEs

### LINKED ISSUE #899: Please update Readme for Speech to Text
service.recognize only support one parameter 
SpeechResults transcript = service.recognize(**audio**, options).execute();

lpatino10:
Fixes #899 

This PR updates the main README as well as the individual service READMEs to make any changes in the examples for the v5.0 release, as well as to make formatting consistent throughout.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ All the services:
 Only Assistant:
 
 ```gradle
-'com.ibm.watson.developer_cloud:assistant:5.0.1'
+'com.ibm.watson.developer_cloud:assistant:5.1.1'
 ```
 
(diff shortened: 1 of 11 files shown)


germanattanasio:
This looks but we need to update the `.bumpconfig` file to make sure the places where the version was not updated will get updated on our next release

lpatino10:
The one place that won't get captured by bumpversion is the SNAPSHOT version in the main README. Unfortunately, after experimenting a bit, I can't find an easy way to add that to our bumpversion configuration file, so we might just have to manually change that when we release.

PR #908, opened 2018-04-02
Update Examples

lpatino10:
This PR updates the example files included in the SDK. This involves:

- Updating the SDK version used in the examples
- Removing examples for deprecated services
- Changing references to `Conversation` to `Assistant`
- Updating any code broken by the SDK version update

These updates should clear up any confusion caused by looking at the examples and trying to use the newest version of the SDK.


germanattanasio:
@lpatino10 Great work here! 👍

PR #914, opened 2018-04-09
Fix bug in the microphone example

flipper83:
### Summary

The microphone example shows the old API.
The continues method doesn't exist anymore. And it's not necessary.
The SpeechResults class has been Renamed to SpeechRecognitionResults.

### Other Information

This is a small change in the readme.


germanattanasio:
@flipper83 can you sign the CLA above?

PR #928, opened 2018-04-17
Apidiff

chohanbin:
This pull request addresses Issue #741 

Produce API diff of the current version of Java-SDK (the version opened in the working directory) and the latest release (published on Maven) by invoking `./apidiff.sh` from the root `java-sdk/`.
Find the produced API diff report in `java-sdk/build/reports/`.

The script that uploads the produced API diffs to WDC GitHub IO page need to be tested with Travis.


apidiff.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
```
lpatino10: Could we name this file to something like `generate-api-diff.sh` to make it slightly more descriptive? I think it would also make more sense to be in the `.utility` folder, which would involve moving directories a little bit in the script.

.utility/push-javadoc-to-gh-pages.sh
```diff
@@ -10,10 +10,16 @@ if [ "$TRAVIS_PULL_REQUEST" == "false" ] && [ "$TRAVIS_BRANCH" ]; then
...
     ../.utility/generate_index_html.sh > index.html
 
+    # produce API diff of the current version and the latest release
+    ./apidiff.sh
+    rm -rf apidiff/
+    mkdir apidiff/
+    cp -f ../build/reports/java-sdk-api-diff* apidiff/
+    ../.utility/generate_apidiff_index_html.sh > apidiff/index.html
+
 	# update the latest/ symlink
```
lpatino10: Weird indentation here

apidiff.sh
```diff
@@ -0,0 +1,53 @@
...
+if [ ! -f $CURRENT_JAR_PATH ]; then
+    echo "apidiff.sh: Current jar was not at its expected location at: $CURRENT_JAR_PATH"
+    exit 1
+fi
+
+# Step 7: Produce an API diff between the latest release and the current version using japicmp module.
+# TODO: Figure out how to set japicmp task's properties (oldClasspath and newClasspath) through command-line invocation.
+./gradlew japicmp -PoldJarPath="${LATEST_RELEASE_JAR_PATH}" -PnewJarPath="${CURRENT_JAR_PATH}" -PoldJarVersion="${LATEST_RELEASE_VERSION}" -PnewJarVersion="${CURRENT_VERSION}"
+
+# Step 8: Construct the filepath to the produced API diff.
```
lpatino10: Do we need these comments? It seems like these steps are happening outside of this script, so it's just a little confusing to have them.

chohanbin: Good catch. I forgot to remove these.

PR #954, opened 2018-06-14
Update authentication section for clarity

SirSpidey:
Addresses changes requested from reviewers to make the authentication section clearer as we migrate services to IAM.

### Summary

- Update intro to authentication
- Move IAM to top of the stack of auth methods. Rename to just `IAM`
- Update getting credentials section
- Update Visual Recognition note
- Change headings to sentence case
- Update toc

cc @mkistler @ehdsouza @anweshan

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,9 +16,9 @@ Java client library to use the [Watson APIs][wdc].
   * [Usage](#usage)
   * [Running in IBM Cloud](#running-in-ibm-cloud)
   * [Authentication](#authentication)
-    * [Username and Password](#username-and-password)
-    * [API Key](#api-key)
-    * [Using IAM](#using-iam)
+    * [IAM](#iam)
+    * [Username and password](#username-and-password)
+    * [API key](#api-key)
   * IBM Watson Services
     * [Assistant](assistant)
     * [Discovery](discovery)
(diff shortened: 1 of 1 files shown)


SirSpidey:
@lopa, one thing missing from IAM here (and most other SDKs) that's available in the [Swift SDK](https://github.com/watson-developer-cloud/swift-sdk/blob/master/README.md#updating-the-accesstoken) is an example of how to refresh the token. Not sure if that's needed in general

```swift
discovery.accessToken("new-accessToken-here")
```

README.md
```diff
@@ -211,13 +191,39 @@ IamOptions options = new IamOptions.Builder()
 service.setIamCredentials(options);
 ```
 
-If at any time you would like to let the SDK take over managing your IAM token, simply override your stored IAM credentials with an IAM API key by calling the `setIamCredentials()` method again.
+### Username and password
+
+```java
```
ehdsouza: python

lpatino10:
What's your opinion on that? We don't offer an exposed method to refresh a user's token, so if their token did expire and they need to use a new one, they would basically need to set the `IamOptions` again like the first time, now with the valid token.

I think that might be fine to leave out of explicit documentation but I could see the case for it.

SirSpidey:
I don't have a strong opinion -- just noting that Mr. Kistler has it. We do have a link in the IAM section to how the process works. Because we're not supporting it, I think it makes more sense not to include it.

PR #977, opened 2018-08-13
Release v6.4.0

lpatino10:
This PR contains changes to be added for the v6.4.0 release. It uses the following commits:
- Codegen: `2590ceea86dc97127f9afa67c79a3c321dd68e5c`
- API definition: `053ee5d76d6bdf82c0cb146a98e459dba2e5b0cd`

Full details will be in the changelog upon release.


discovery/src/main/java/com/ibm/watson/developer_cloud/discovery/v1/model/CreateEnvironmentOptions.java
```diff
@@ -20,17 +20,41 @@
...
+    /** XXL. */
+    String XXL = "XXL";
+    /** XXXL. */
+    String XXXL = "XXXL";
+  }
+
   private String name;
   private String description;
-  private Long size;
+  private String size;
```
dpopp07: I missed the Friday meeting, so I may have heard wrong, but wasn't the decision to leave `Size` in the SDK as a number for compatibility and add a `SizeString` parameter to be type `String`?

lpatino10: You might be right @dpopp07, although then that'll involve us having more handwritten methods for backwards-compatibility in the next major release when the `Long` size is the only one left.

dpopp07: This is for the request, not the response, right? I know it's been deprecated for a long time, so users shouldn't be including it in their request, but if they are it would probably break them. If you think this is okay and that we shouldn't be causing any issues, I'd prefer to do the same thing in the Node SDK and just change the type. So let me know!

germanattanasio: You can take the same approach as the one logan is taking here.

dpopp07: @lpatino10 I agree, which is why I was I wasn't thrilled about doing it haha
@germanattanasio Sounds good, I'll make that change then

PR #981, opened 2018-08-31
build(appscan): Add scripts for Appscan in Travis on release

mediumTaj:
### Summary
This pull request adds AppScan to the Travis build on tagged release.

If approved, please squash and merge this commit with the message
```
build(appscan): Add scripts for Appscan in Travis on release
```


germanattanasio:
How hard will it be to move all this into an npm package. You can do `npm install travis-app-scan` or something like like that. Looks like most of the code will be the same across SDKs.

mediumTaj:
it was a little different for the two that I've done so far - Go has a particular go directory it installs to and for some reason this java build didn't like it when I tried to change to 32 bit to run the ASOC client. I can look into this though.

mediumTaj:
After speaking with @dpopp07 we may be able to convert the makefile into an npm module but we need to find the differences of implementing this solution into each SDK first. I will make some changes here to move all appscan related files into a separate directory and continue to implement the current solution across the SDKs. @dpopp07 can help turn this into an npm module once we find the differences across SDKs.

PR #994, opened 2018-10-12
Added StringHelper.createFormPartName and tests

padamstx:
### Summary

The new StringHelper.createFormPartName() function introduced by this PR will be used by generated java code when the "mapified form parameter" feature is used.

### DIFF
--- a/core/src/main/java/com/ibm/watson/developer_cloud/util/StringHelper.java
+++ b/core/src/main/java/com/ibm/watson/developer_cloud/util/StringHelper.java
@@ -33,4 +33,22 @@ public class StringHelper {
         // TODO - check to see if we need to be concerned with the encoding of the string!
         return new ByteArrayInputStream(s.getBytes());
     }
+
+    /**
+     * Creates a parameterized part name for a form part based on a template and a key name and value.
+     * Example:
+     *     template: "{classname}_positive_examples"
+     *     keyName:  "classname"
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/ibm/watson/developer_cloud/util/StringHelper.java
```diff
@@ -33,4 +33,22 @@ public static InputStream toInputStream(String s) {
...
+     *
+     * @param template the part name template
+     * @param keyName the name of the key within the template
+     * @param keyValue the value of the key to use as a replacement
+     * @return
+     */
+    public static String createFormPartName(String template, String keyName, String keyValue) {
+        String toBeReplaced = "{" + keyName + "}";
+        return template.replace(toBeReplaced, keyValue);
+    }
```
mkistler: It seems like we could avoid having this special function if we did just a little more work in the generator.

In particular, I think we could, in the generator, create a "format string" from the template:
```
"{classname}_positive_examples" -> "%s_positive_examples"
```

Then in the service method, we change
```
String partName = StringHelper.createFormPartName("{{partNameTemplate}}", "{{mapKeyName}}", entry.getKey());
```
to
```
String partName = String.format("{{partNameFormatString}}", entry.getKey());
```

padamstx:
This PR will not be needed afterall...

PR #1006, opened 2018-11-14
Fix Speech to Text bugs

### LINKED ISSUE #1004: [speech-to-text] EOFException during audio file processing
I am developing application that utilizes IBM speech to text feature. For that purpose I am using Java API. For some files `BaseRecognizeCallback` interface is called with `onError` method, with `java.io.EOFException` with message set to `null`. This happens when I call speech to text with some larger files.
 
- [x] Steps to reproduce:
In order to reproduce this issue, I used your example project, `RecognizeUsingWebSocketsExample` class and I commented out `interimResults` (since I don't need it) and run the example with my wav file, given on the link https://drive.google.com/drive/u/0/folders/1wqnn_svof8FrR4jmygQKBo131vw00fr6?ogsrc=32
- [x] Expected behavior
IBM should return valid `SpeechRecognitionResults`
- [x] Actual behavior
`onError` method is called with `java.io.EOFException` with message set to `null`.
- [x] JDK version
10.0.2
- [x] java-sdk version
6.9.1

### LINKED ISSUE #1005: [speech-to-text] Interim results set to true but seeing final results
I have set the .interimResults(true) for Speech to Text new RecognizeOptions.Builder()
but still seeing the final results than interim results

When reporting a bug, please be sure to include the following:

I have used the examples code and set the interimResults to true. But when I check the response I always see this

{
      "result_index": 0,
      "results": [
        {
          "alternatives": [
            {
              "confidence": 0.728,
              "transcript": "hi what should I call you "
            }
          ],
          "final": true
        }
      ]
    }

Just figured out that the getRecognizeOptions is never setting the interimResults. getRecognizeOptions output

 {
      "content-type": "audio/ogg;codec=opus",
      "inactivityTimeout": 2000,
      "model": "en-US_BroadbandModel"
    }

After looking into the code, I think maybe because https://github.com/watson-developer-cloud/java-sdk/blob/master/speech-to-text/src/main/java/com/ibm/watson/developer_cloud/speech_to_text/v1/model/RecognizeOptions.java#L441 is missing `interimResults = builder.interimResults`

lpatino10:
Fixes #1004 and fixes #1005.

A ping interval was added in the `recognizeUsingWebSocket()` method to ensure that no timeouts occur. Apparently, this was causing issue #1004, according to [this issue](https://github.com/square/okhttp/issues/3113) in the OkHttp repo.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 6.9.1
+current_version = 6.9.2
 commit = True
 message = docs: Update version numbers from {current_version} -> {new_version}
 
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ All the services:
 <dependency>
 	<groupId>com.ibm.watson.developer_cloud</groupId>
 	<artifactId>java-sdk</artifactId>
-	<version>6.9.1</version>
+	<version>6.9.2</version>
 </dependency>
 ```
 
@@ -70,7 +70,7 @@ Only Discovery:
 <dependency>
 	<groupId>com.ibm.watson.developer_cloud</groupId>
 	<artifactId>discovery</artifactId>
-	<version>6.9.1</version>
+	<version>6.9.2</version>
 </dependency>
 ```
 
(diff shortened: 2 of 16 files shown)


lpatino10:
Nope @mediumTaj, that should override the pings to close the connection. Good question though!

PR #1015, opened 2018-12-07
Add WebSocket support for Text to Speech synthesize endpoint

lpatino10:
This PR adds support for WebSockets in the Text to Speech `synthesize()` method. This new method is called `synthesizeUsingWebSocket()`, and should very closely mirror the `recognize()` and `recognizeUsingWebSocket()` methods in Speech to Text.

Integration tests have been added to ensure that this is working properly, as well as an example in the Text to Speech README for new users.


lpatino10:
@germanattanasio can you review again? All I did was fix those checkstyle issues.

PR #1025, opened 2019-01-03
CreateDialogNode fix

lpatino10:
This PR adds a missing field to the `CreateDialogNode` model in both the Assistant v1 and Conversation services. This fix should be automated for the next release.


lpatino10:
Looks like STT is having some timeout issues that's messing with CI 😞

PR #1027, opened 2019-01-10
Support credential file

lpatino10:
This PR adds SDK support for the credential file format that will soon be downloadable on IBM Cloud. With this change, the SDK will automatically search for a file called `ibm-credentials.env` when a service class is constructed. The following locations are searched, in this order:

- System home directory (Unix)
- System home directory (Windows)
- Top-level of current project directory

Users also have the option of setting an environment variable called `IBM_CREDENTIALS_FILE` with the location and name of their credential file if they'd like to customize it. This file will take precedence over the above files.


lpatino10:
Leaving as a work in progress for now to get feedback until the file is actually supported on the dashboard.

mkistler:
I don't know that it is meaningful / useful to refer to the "User's home directory" as a search location for the credentials file.  Consider an application that is deployed into a Kubernetes cluster and uses the Java SDK to interact with some Watson service -- who is "the user" is this case?

lpatino10:
Good point @mkistler...maybe "system home directory" is more accurate.

lpatino10:
@jsstylos mentioned that we might want to add support for other types of values like the `assistant_id` in Assistant v2 to have feature parity with current demos like [this one](https://github.com/watson-developer-cloud/assistant-simple).

germanattanasio:
Can we add an environment variable like `IBM_CREDENTIALS_FILE` so that users can add the value in that variable rather than hardcoding it in the code?

lpatino10:
@germanattanasio @mediumTaj I just added the support for the environment variable and updated the README if you'd like to take a look. I think it should be ready to look at.

Regarding Jeff's comment about adding other types of values, I think we can hold off on that for a later addition and that this is enough for an MVP.

lpatino10:
@mediumTaj Right now it won't support multiple instances of the same service, so that'll also have to be a future improvement.

PR #1031, opened 2019-01-21
Fix credentials bug

amanoese:
### Summary

Fix bugs when different authentication of each instance of same service if there is vcapserices environment.

This bug happend my cloudfoundry app when bind two discovery servicies.
I need this patch until complete of migration authentication from old to iam.


amanoese:
missing commit log mail address.

PR #1036, opened 2019-01-31
Add headers for analytics

lpatino10:
This PR adds a header for each method call which sends analytics information about the service, version, and endpoint called. It also tweaks the format of the `User-Agent` header to match that of the other SDKs, with one small difference. Here's an example of the new `User-Agent`:

`watson-apis-java-sdk-6.13.2 Mac OS X 10.14.1 Oracle Corporation 1.8.0_144`

The Java vendor is added before the Java version (both the last values in the string), because I think this will also help guide us in terms of what vendors to support.


lpatino10:
@ehdsouza would you be able to take a quick look again? I had to fix a test and your old review went stale.

PR #1043, opened 2019-03-07
Add common project

lpatino10:
This PR adds a new `common` project with an `SdkCommon` class to handle logic that's common to the services, but doesn't belong in the core since it's project-specific. Currently, that means it handles building the `User-Agent` and `X-IBMCloud-SDK-Analytics` values that will be added to each service request.

This PR won't contain the changes to the service methods, but for reference, the code will look something like this:
```java
public ServiceCall<QueryResponse> query(QueryOptions queryOptions) {
  ...
  RequestBuilder builder = RequestBuilder.post(RequestBuilder.constructHttpUrl(getEndPoint(), pathSegments, pathParameters));
  builder.query(VERSION, versionDate);
  if (queryOptions.loggingOptOut() != null) {
    builder.header("X-Watson-Logging-Opt-Out", queryOptions.loggingOptOut());
  }

  // Add any SDK-specific "default" headers.
  Map<String, String> defaultHeaders = SdkCommon.getDefaultHeaders("discovery", "v1", "query");
  for (Entry<String, String> header : defaultHeaders.entrySet()) {
    builder.header(header.getKey(), header.getValue());
  }
  ...
  final JsonObject contentJson = new JsonObject();
  ...
  builder.bodyJson(contentJson);
  return createServiceCall(builder.build(), ResponseConverterUtils.getObject(QueryResponse.class));
}
```

### DIFF
--- a/common/src/main/java/com/ibm/watson/common/SdkCommon.java
+++ b/common/src/main/java/com/ibm/watson/common/SdkCommon.java
@@ -14,6 +14,8 @@ public class SdkCommon {
   private static final Logger LOG = Logger.getLogger(SdkCommon.class.getName());
   private static String userAgent;
 
+  private SdkCommon() { }
+
   private static String loadSdkVersion() {
     ClassLoader classLoader = SdkCommon.class.getClassLoader();
     InputStream inputStream = classLoader.getResourceAsStream("java-sdk-version.properties");
--- a/common/src/main/java/com/ibm/watson/common/SdkCommon.java
+++ b/common/src/main/java/com/ibm/watson/common/SdkCommon.java
@@ -1,0 +1,48 @@
+package com.ibm.watson.common;
+
+import com.ibm.cloud.sdk.core.http.HttpHeaders;
+import com.ibm.cloud.sdk.core.util.RequestUtils;
+
+import java.io.InputStream;
+import java.util.HashMap;
+import java.util.Map;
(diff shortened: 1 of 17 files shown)


lpatino10:
Leaving this as a draft until the core library is updated from [this PR](https://github.com/IBM/java-sdk-core/pull/12). Then, I'll update dependencies on this side to get everything working.

common/src/main/java/com/ibm/watson/common/SdkCommon.java
```diff
@@ -0,0 +1,48 @@
...
+
+    String sdkAnalyticsHeaderValue = String.format(
+        "service_name=%s;service_version=%s;operation_id=%s",
+        serviceName,
+        serviceVersion,
+        operationId
+    );
+
+    headers.put(HttpHeaders.X_IBMCLOUD_SDK_ANALYTICS, sdkAnalyticsHeaderValue);
+    headers.put(HttpHeaders.USER_AGENT, getUserAgent());
```
padamstx: technically speaking... :)  the user agent header value could be stored in a static field and not computed each time getDefaultHeaders() is called.   I'm sure the performance difference is negligible, but I couldn't stop myself from mentioning it :)

lpatino10: Any performance improvement is a good one! Done

PR #1066, opened 2019-06-10
Update readme to change version to `7.x.x` and link directly to releases page

mediumTaj:
### Summary
This pull request updates the version listed in the readme to `7.x.x` and links directly to the releases page for release notes. I'm not sure if we want to say `7.x.x` here or the current version, `7.1.0`. Maybe we can use bumpversion for this?


lpatino10:
I think this section is a bit redundant with the major release past us. I'm going to go ahead and close this and remove that bit entirely.

PR #1068, opened 2019-06-15
Fix AudioMetrics bug (Speech To Text)

TakahiroSakuda:
### Summary

With WebSocket, AudioMetrics is returned as a other response.
AudioMetrics was ignored because neither "results" nor "speaker_labels" were included.


TakahiroSakuda:
Thanks!

PR #1081, opened 2019-08-21
Update Compare and Comply with model fix

lpatino10:
This PR updates Compare and Comply to fix what was a broken representation of the `BodyCells` model. Additionally, a new model has been added to the `ClassifyReturn` response. Tests have been added to cover both changes.


lpatino10:
Looks like there are currently some DIscovery service issues, hence the CI failure 😕  I'll rebuild again once that seems like it's sorted out.

Issue #1095, opened 2019-11-06
Jar with dependencies contains unnecessary and potentially harmful dependencies


smccants:
**Overview**
I'm trying to use the ibm-watson-8.0.0-jar-with-dependencies.jar in my application and cannot because it contains unnecessary class files that conflict with classes included in my application.

**Expected behavior**
Be able to add the jar file to my application and successfully compile.

**Actual behavior**

```
:app:transformClassesWithDexBuilderForDebug
 AGPBI: {"kind":"error","text":"Program type already present: edu.umd.cs.findbugs.annotations.CreatesObligation","sources":[{}],"tool":"D8"}
 :app:transformDexArchiveWithExternalLibsDexMergerForDebug FAILED
 
 FAILURE: Build failed with an exception.
 
 * What went wrong:
 Execution failed for task ':app:transformDexArchiveWithExternalLibsDexMergerForDebug'.
 .....
   Program type already present: edu.umd.cs.findbugs.annotations.CreatesObligation
   Learn how to resolve the issue at https://developer.android.com/studio/build/dependencies#duplicate_classes.
```

**How to reproduce**
Build an application that includes FindBugs and then include the java-sdk jar file.  The two FindBugs classes will conflict with each other,

**SDK Version**
8.0.0

**Additional information:**
- OS: Android 7.x+, built under KUbuntu 14.04 LTS
- Which version of `Java` are you using? 1.8

**Additional context**
I think the problem is a variety of "dependencies" are included in the jar file that are unnecessary and potentially harmful.  I had to remove the following from the jar in order to get my application to compile:
- Findbugs
- javax.annotations
- net/jcip/annotations

I also think (but I'm not sure) the following could be removed:
- org.checkerframework
- org.codehaus
- org.hamcrest
- org.mockito
- org.objenesis
- org.reactivestreams

smccants:
Switching from using the jar file to the gradle include

    implementation 'com.ibm.watson:ibm-watson:8.0.0'

Results in the same problem:

```
:app:transformClassesWithDexBuilderForDebug
AGPBI: {"kind":"error","text":"Program type already present: edu.umd.cs.findbugs.annotations.CreatesObligation","sources":[{}],"tool":"D8"}
:app:transformDexArchiveWithExternalLibsDexMergerForDebug FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:transformDexArchiveWithExternalLibsDexMergerForDebug'.
[...]

PR #1096, opened 2019-11-11
Update core package

lpatino10:
This PR updates the core SDK package, mainly to fix https://github.com/watson-developer-cloud/java-sdk/issues/1095. Other core changes included in this update can be found here: https://github.com/IBM/java-sdk-core/releases/tag/7.1.0

Besides the core update, I reconfigured the automated release to fix an issue that had been occurring where bumpversion commits were being excluded from `master`.


.releaserc
```diff
@@ -1,17 +1,19 @@
 {
   "branch": "master",
   "tagFormat": "java-sdk-${version}",
+  "debug": true,
```
padamstx: technically this doesn't need to be here... I think the debug flag is really just used to produce more verbose output in the travis build log (possibly... presumably...  :) ).

.releaserc
```diff
@@ -1,17 +1,19 @@
...
   "plugins": [
+    "@semantic-release/changelog",
```
padamstx: In the go and python cores, I didn't specify a "plugins" field.  Instead, the list of plugins are listed in the "prepare" field.
I don't know what the difference is, but I do know that those two repos seems to be working correctly now :)
But the key is to ensure that the "git" plugin is executed last I think after the steps that create commits are finished.
Here's a link to the go core .releaserc file in case you want to take a look at that for comparison purposes:
https://github.com/IBM/go-sdk-core/blob/master/.releaserc

assistant/build.gradle
```diff
@@ -60,7 +60,7 @@ checkstyle {
 dependencies {
     compile project(':common')
     testCompile project(':common').sourceSets.test.output
-    compile 'com.ibm.cloud:sdk-core:7.0.0'
+    compile 'com.ibm.cloud:sdk-core:7.1.0'
```
padamstx: If you end up using the latest version of the generator to generate the upcoming Watson sdk release, you'll need to upgrade to the new major version of the java core (8.0.0) after that is tagged (should be soon).   Just a heads up.

.releaserc
lpatino10: Yep, I went ahead and added it _just_ in case this change still isn't it so that I can fix it easier 😬

.releaserc
lpatino10: Yeah I saw that you were doing it a bit differently. The only reason I went with the "plugins" approach is that it seems to be the new standard way to define it, at least according to semantic-release documentation: https://github.com/semantic-release/semantic-release/blob/master/docs/usage/configuration.md#options

Since we're running `npx semantic-release`, we should be grabbing the latest version. I'm just a bit worried that some build in the future might grab a major release and remove support for what seems to be the old way of defining things. I tried to still re-arrange the order to match what you've done since that was presumably the issue.

PR #1113, opened 2020-03-06
fix(Discovery v1): environment document count should use 'available'

gjs29:
### Summary

The Discovery service returns 'available' as the total number of documents for an environment.
This was previously using 'indexed' which is not what the service returns.


mediumTaj:
@gjs29 Can you prefix the commit message with `fix(Discovery v1): ` so that CI will do a patch release of the Java SDK?

ie:
```
fix(Discovery v1): Discovery environment document count should use 'available'
```

mediumTaj:
This does not constitute a breaking change since the v1 api never returned this as `indexed` in the first place. This has been addressed in the API definition and should be picked up in future regeneration cycles.

PR #1118, opened 2020-04-09
build: switch to travis stages

germanattanasio:
Switch to use travis stages in order to prevent integration tests from failing the build


assistant/src/test/java/com/ibm/watson/assistant/v1/AssistantServiceIT.java
```diff
@@ -16,7 +16,7 @@
 import com.ibm.cloud.sdk.core.http.ServiceCallback;
 import com.ibm.cloud.sdk.core.security.BasicAuthenticator;
 import com.ibm.cloud.sdk.core.service.exception.NotFoundException;
-import com.ibm.cloud.sdk.core.service.exception.UnauthorizedException;
+import com.ibm.cloud.sdk.core.service.exception.ForbiddenException;
```
mamoonraja, line 19: Noticed this is only changed in `Discovery` and `Assistant`. Confirming it that's all where we need this change.

germanattanasio: Yeah. it's because of the IAM change they did

PR #1134, opened 2020-10-14
Fix missing option assignments

JxSun:
The `RecognizeOptions` constructor doesn't really apply the `interimResults`, so the client app can't receive non-final results as stated in the document.
This patch adds the missing `interimResults` assignment (and also the missing `processingMetrics` and `processingMetricsInterval`) back to the constructor.


kevinkowa (approved):
Looks good to me! Good catch, thank you!

PR #1136, opened 2020-10-23
fix(java-sdk-core-version): update java-sdk-core version to fix seria…

kevinkowa:
update java-sdk-core version to fix serialization error on disco-v2 with QueryResponse model in the toString() method.


mediumTaj (approved):
Looks good after CI

PR #1140, opened 2020-11-17
Update visual rec

kevinkowa:
We should **not** have trainingStatus as a parameter for VisRec V4 CreateCollection and UpdateCollection operations. This field is marked as readonly


germanattanasio (approved):
LGTM

PR #1157, opened 2021-04-07
Assistant next step

kevinkowa:
### Summary

This is a hand written change to update next_step in a dialog node to null.

### DIFF
--- a/assistant/src/main/java/com/ibm/watson/assistant/v1/Assistant.java
+++ b/assistant/src/main/java/com/ibm/watson/assistant/v1/Assistant.java
@@ -2248,6 +2248,12 @@ public class Assistant extends BaseService {
           com.ibm.cloud.sdk.core.util.GsonSingleton.getGson()
               .toJsonTree(updateDialogNodeOptions.newNextStep()));
     }
+    if (updateDialogNodeOptions.deleteNextStep() != null && updateDialogNodeOptions.deleteNextStep()) {
+      contentJson.add(
+              "next_step",
+              com.ibm.cloud.sdk.core.util.GsonSingleton.getGson()
+                      .toJsonTree(null));
+    }
     if (updateDialogNodeOptions.newTitle() != null) {
       contentJson.addProperty("title", updateDialogNodeOptions.newTitle());
     }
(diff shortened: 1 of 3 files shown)


apaparazzi0329:
This is looking good. I would add an integration test to ensure that a previously populated `next_step` is now empty after applying the new parameter

apaparazzi0329 (approved):
LGTM

apaparazzi0329:
My only concern right now is how are we going to document this for other users since it will not be on the official Assistant docs

apaparazzi0329:
Maybe add a more descriptive comment for the parameter and what it does

assistant/src/main/java/com/ibm/watson/assistant/v1/Assistant.java
```diff
@@ -2248,6 +2248,12 @@ public void setVersion(final String version) {
           com.ibm.cloud.sdk.core.util.GsonSingleton.getGson()
               .toJsonTree(updateDialogNodeOptions.newNextStep()));
     }
+    if (updateDialogNodeOptions.deleteNextStep() != null && updateDialogNodeOptions.deleteNextStep()) {
+      contentJson.add(
+              "next_step",
+              com.ibm.cloud.sdk.core.util.GsonSingleton.getGson()
+                      .toJsonTree(null));
```
mediumTaj, line 2255: Does this actually put `null` into `next_step` in the json? want to make sure gson isn't disregarding null values

assistant/src/main/java/com/ibm/watson/assistant/v1/model/UpdateDialogNodeOptions.java
```diff
@@ -731,4 +747,15 @@ public Boolean newDisambiguationOptOut() {
...
+  /**
+   * Gets the deleteNextStep.
+   *
+   * <p>Set nextStep to null
+   *
+   * @return the deleteNextStep
+   */
+  public Boolean deleteNextStep() {
+    return deleteNextStep;
+  }
```
mediumTaj, line 760: this is a lot of handwritten changes

mediumTaj (commented):
Lets talk about this PR during parking lot.

assistant/src/main/java/com/ibm/watson/assistant/v1/Assistant.java
kevinkowa: yes, this is the output from `nc -l` 
```
{"dialog_node":"Test2d28f9c6e-8864-4bbb-8ab8-81d18b740ce9","description":"Updated description of Test79155c70-05aa-47a3-b1e2-5557109f89b3","next_step":null}
```

apaparazzi0329: It does. We looked at the final request sent to the service and it `inserts` null into the request

PR #1181, opened 2021-08-16
fix(pom.xml): update java-sdk-core version

kevinkowa:
Updating the version of the core to `9.12.1` so we can allow users to use username and apikey auth for cpd


mediumTaj (approved):

repjarms (approved):
LGTM

PR #1186, opened 2021-08-23
ci: 9662 addcheck

Mikemosca:
### Summary

I removed the tag build check which did not seem to work for GHA  and put in what we do for dotnet.

1) Do a semantic release in dry run and see if it will generate a release
2) If release is true do a real semantic release and release to maven
3) if release is false print message release is skipped


kevinkowa (approved):
LGTM!

PR #1187, opened 2021-08-31
ci: Gha gpg

Mikemosca:
### Summary

many fixes for gpg and maven central


kevinkowa (approved):
LGTM, but we might need to squash the commits

PR #1196, opened 2021-11-22
ci: Gha 9907

Mikemosca:
Add integration test workflow


kevinkowa (commented):
is there an issue with these System.out.println ?

Mikemosca:
> 



> is there an issue with these System.out.println ?

No.  I just commented them out because they add lots of output to the results stream.  I can put them back if they add value.  Up to you,

kevinkowa:
> > 
> 
> > is there an issue with these System.out.println ?
> 
> No. I just commented them out because they add lots of output to the results stream. I can put them back if they add value. Up to you,

That's fine, just wondering

kevinkowa (approved):
lgtm!

PR #1198, opened 2021-12-01
fix(deploy.yml): up the version for node to fix semantic-release

kevinkowa:
semantic-release failed since the requirement for it is now node 14. This PR upgrades the node version to fix this issue.


Mikemosca (approved):
LGTM, hope that should do it

PR #1209, opened 2022-04-25
 fix(language-translator-v3): add content-type to createModelOptions model properties

kevinkowa:
### Summary

Add content-type to properties that require a file within the createModelOptions class


apaparazzi0329 (approved):
LGTM

PR #1220, opened 2023-08-03
fix(pom): upgrade sdk-core

apaparazzi0329:
This upgrade is mainly to remove the vulnerable guava dependency


jeff-arn:
@apaparazzi0329 can you verify that this dependency update isn't introducing compilation errors in the SDK? It looks like the build and test action is failing to compile.

apaparazzi0329:
Yes I will double check everything. I did not have any issues running tests locally

PR #1228, opened 2024-05-15
Major SDK Release 13.0.0

apaparazzi0329:


speech-to-text/src/main/java/com/ibm/watson/speech_to_text/v1/model/RecognizeWithWebsocketsOptions.java
```diff
@@ -234,9 +232,7 @@ public static class Builder {
     private Boolean splitTranscriptAtPhraseEnd;
     private Float speechDetectorSensitivity;
     private Float backgroundAudioSuppression;
-    private Boolean lowLatency;
     private Float characterInsertionBias;
-    private Boolean interimResults;
```
kevinkowa, line 239: Wouldn't that be considered a breaking change?

kevinkowa (approved):
LGTM, just a comment if that is a breaking change and willing to introduce it into a minor release

speech-to-text/src/main/java/com/ibm/watson/speech_to_text/v1/model/RecognizeWithWebsocketsOptions.java
apaparazzi0329: You are right. For the java-sdk this is a breaking change. I hate to have a major release for such a small change but it would be necessary

kevinkowa (approved):
lgtm!

